PR #1, opened 2019-11-25
fix typo

radekosmulski:


sgugger:
Thanks for your PR. You need to make the change in `index.ipynb` since the README is auto-generated from it. I'll port this one, just warning for next time :).
Will also try to add a comment to markdown source but not source how to do that.

PR #4, opened 2019-11-29
added dockerfile

hamelsmu:


hamelsmu:
Ooops sorry, meant to test locally first will open a PR later

PR #6, opened 2019-11-29
Docker build

hamelsmu:
cc: @jph00 following up from our conversation this PR does the following:

- Uses Actions to build and push a docker container to DockerHub.  
- Edits the CI workflow to use the Docker container instead of building the environment in the workflow.  The advantage is for other people, they can use nbdev in their Action workflow without the overhead of building each time.

For this to work, you must

1. Create a DockerHub repo called `fastai/nbdev`
2. Save your username & password
3. Add the following [secrets to GitHub](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/creating-and-using-encrypted-secrets) (1) `DOCKER_USERNAME` and (2) `DOCKER_PASSWORD`

- This PR edits the original `main.yml` to use the Docker container.  The way this file is setup, we will build the Docker container each time, but other people who use your template can instead do something like this.  I will make a PR in the template repo too

```yaml
name: CI
on: [push]

jobs:
  nbdev:
    name: Run nbdev checks
    needs: build_container
    runs-on: ubuntu-latest
    # use the container to run the checks
    container: fastai/nbdev
    steps:
    - uses: actions/checkout@v1
    - name: Read all notebooks
      run: |
        nbdev_read_nbs
    - name: Check if all notebooks are cleaned
      run: |
        echo "Check we are starting with clean git checkout"
        if [ -n "$(git status -uno -s)" ]; then echo "git status is not clean"; false; fi
        echo "Trying to strip out notebooks"
        nbdev_clean_nbs
        echo "Check that strip out was unnecessary"
        git status -s # display the status to see which nbs need cleaning up
        if [ -n "$(git status -uno -s)" ]; then echo -e "!!! Detected unstripped out notebooks\n!!!Remember to run tools/run-after-git-clone"; false; fi
    - name: Check if there is no diff library/notebooks
      run: |
        if [-n "$(nbdev_diff_nbs)"]; then echo -e "!!! Detected difference between the notebooks and the library"; false; fi
    - name: Run tests
      run: |
        nbdev_test_nbs
```


hamelsmu:
I made a corresponding PR in https://github.com/fastai/nbdev_template/pull/1 to reflect these changes

jph00:
Closing after discussion with @hamelsmu since we want to avoid users having a hard-coded py version.

aronchick:
Interesting question - what if we did this as a docker container with a hard coded python version? E.g. https://github.com/sethvargo/go-githubactions

jph00:
I'm not sure the upside (about 15 seconds saved in CI time) is worth the downside (both complexity, and challenges in integrating docker configuration with nbdev configuration).

PR #7, opened 2019-11-29
Tools

holyfiddlex:
Tools for notebook development, Mostly taken from the fastai_dev library but seem like good tools to have available when working on nbs. So far the tools are just the ones I used the most, store_attr, patches and delegates.

Still not finished, still want to make changes to allow chained decorators to work, change patching to work on classmethods, setters and getters and a delegate_args function. Also missing documenation, mostly checking to see if this is something the creators would be interested in adding or not.


sgugger:
The core of fastai v2 functionality (basically everything that is in fastai2.core) will be packaged in a separate repo. We'll tackle this once we've fully released nbdev, so there is no need pursuing this draft since we want nbdev to be just about exporting/testing/docs.

PR #11, opened 2019-12-03
small typo fix

Zymrael:
Minor fix in the README. Thanks for the great repo btw, jupyter only dev environments feel much more efficient now.


sgugger:
Thanks for your PR. As mentioned at the beginning of the README however, this typo fix is going to be useless since the README file is autogenerated from the idnex notebook. So next time we build the docs, it will be erased. I'll port your fix to that notebook but for next time, remember to change index.ipynb then run `make docs` :)

PR #14, opened 2019-12-04
Support of LaTeX in GitHub Pages added

kdorichev:
As suggested: [https://stackoverflow.com/a/46765337/6551233](https://stackoverflow.com/a/46765337/6551233)
Example of the result: [https://kdorichev.github.io/deeplearning/core](https://kdorichev.github.io/deeplearning/core)


jph00:
Thanks for the suggestion! I've actually added KaTeX to nbdev_template now since I've found that's a lot faster than mathjax.

PR #16, opened 2019-12-05
minor documentation updates 

gkennos:
* addition to getting started that is helpful if someone chooses to create new .ipynb instead of modifying provided sample 
* suggestion to make it clear the best way to import your generated module if storing notebooks anywhere other than project root, without requiring sys.path hacks


sgugger:
Thanks! I'll also add the solution to create a simlink when you have a path for the notebooks that's not just the origin.

PR #23, opened 2019-12-20
Improved Pep8 Compability for Autogenerated Code

penguinmenac3:
## Motivation

Even though this is a tiny change, for teams and single developers working on large projects, where pep8 and some other linting tool satisfaction is required for commit and push, this is a game changer, whereas the developer has control over the content of the cell, but not on how the cells are converted in auto generated code.

I do not want to make nbdev pep8 compatible, since I suppose you probably don't care for a reason, but this fix empowers the users who want to write pep8 compatible code with nbdev to do so.

## Issues Addressed

This change ensures that there are two empty lines between every cell. Pep8 requires to have two lines before  a function call. The code from the cells itself is not modified, only the auto generation process. ("PEP 8: expected 2 blank lines, found 1")

To further improve the "pep8 correctness" of auto generated code, the generated comments contain a space between the hash and the comment. ("PEP 8: block comment should start with '# '")


sgugger:
Thanks for your PR! After discussing it with Jeremy, we will merge this PR and amend it a little bit:
- the space after the # is fine
- the two lines between any cell is a bit more annoying for users would don't really care about PEP8 so we will make this configurable with an int (default to 1) in the settings.ini file. You can then set it to 2 to conform to PEP8.

I will ensure this doesn't impact the conversion back and forth or the diff scrip, every should be finished by the end of today.

Thanks again!

PR #27, opened 2019-12-23
Add permalink: pretty as default option in the _config.yml template.

daviddemeij:
When trying to deploy the nbdev documentation to an authenticated heroku app (using https://github.com/benbalter/jekyll-auth) it turned out that the extensionless URLs in the menu didn't work. When I manually added `.html` to the end of the URL I did find the page.

This was easily fixed (although it took some time to find this solution) by adding a parameter `permalink: pretty` to `_config.yml`.

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -56,6 +56,7 @@ defaults:
       tooltip: true
 
 sidebars: [home_sidebar]
+permalink: pretty
 
 #theme: jekyll-theme-cayman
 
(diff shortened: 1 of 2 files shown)


sgugger:
This file is auto-generated from the [template module](https://github.com/fastai/nbdev/blob/master/nbdev/template.py) so you should add the same there (line 107) otherwise it won't be effective on new modules.

daviddemeij:
I was looking for something like that but I couldn't find it so I thought it was generated from the config in the docs somehow, thanks for the tip! I added it to that file as well now.

sgugger:
Thanks a lot!

PR #31, opened 2019-12-31
Update autogenerated text to use relative path between lib code and nbs

ltirrell:
Previously, running `notebook2script` created autogenerated text saying that the notebook used to generate library code was in `f'dev/{nb_name}'`. Now, the relative path between the code and the ipynb is used.

I added one test, showing that the autogenerated line for `export.py` is as expected based off the `settings.ini`, let me know if you would like more (I changed the `nbs_path` and `lib_path` to a bunch of different values when initially testing and it also worked, but didn't include these as additional files were written).


sgugger:
Thanks a lot! the dev was legacy code, having the relative path is better.

PR #36, opened 2020-01-07
add option to format docstring as monospace

maarten990:
In our existing codebase, we like having detailed doc strings. With the current behaviour, documentation is pretty but docstrings lose their newlines/indentation, which is a problem for most of the common ways of formatting arguments. This adds a `monospace_docstrings` config option that adds `<pre>` tags around the doc strings.

Screenshot 2020-01-07 at 12 35 20

### DIFF
--- a/docs/showdoc.html
+++ b/docs/showdoc.html
@@ -1380,7 +1380,7 @@ summary: "Functions to show the doc cells in notebooks"
 </div>
 <div class="cell border-box-sizing text_cell rendered"><div class="inner_cell">
 <div class="text_cell_render border-box-sizing rendered_html">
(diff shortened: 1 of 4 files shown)


sgugger:
I'll look a this later today. It seems interesting! Out of curiosity, why don't the doc links work in the multi-line version? I sit a bug on our side?

maarten990:
When I add doc links, the example from the screenshot would end up looking like
```
Doc link: [show_doc](whatever_the_link_would_be)
Args:
    ... etc
```
so with an uninterpreted Markdown syntax link in there.
Perhaps the links would work if they were translated from Markdown to HTML before ending up inside the `<pre>` tags, but that was a little bit more effort than I was willing to do before getting any feedback on the idea 😅 
I'm also not much of an HTML guy, so take this with a grain of salt.

sgugger:
Ok, merging in this state and putting that on my TODO for when I have time. Thanks for the PR!

Issue #40, opened 2020-01-09
function reference URL in tip or note will break jekyll build


daviddemeij:
I discovered that when using a function reference URL in a `> note:` or `> tip:` it will break the jekyll build locally as well as on my Heroku app with the following exception:
```
Liquid Exception: Invalid syntax for include tag: content="The <a href="utils#example_func">The <code>example_func</code></a> does something ." Valid syntax: {% include file.ext param='value' param2='value' %} in utils.html
```
I confirmed this by removing one of the links (then it gave an error on the next one) and by removing both of them (solving the error).

I think this is caused by the double quotes in the url breaking the `content=" ... "`, maybe a solution would be to use single quotes in the translation of the urls to html. I will try this and make a pull request if it works.

sgugger:
Fixed by #41

PR #41, opened 2020-01-09
Fix jekyll note doclinks

daviddemeij:
Changed the double quotes in the `add_jekyll_note` function by single quotes escaped (using `\`). I also updated the test case and added an example of a doc link in a Jekyll note to show that it works now.

Tested this for the nbdev docs as well as for my own docs (where I initially encountered the problem) and it seems to solve the problem in both cases.

Solves #40


daviddemeij:
To keep it clean, I did not rebuild the entire docs yet since it seems some other things also changed (that are unrelated to my changes).

sgugger:
Thanks, I'll rebuild the doc as a follow-up commit!

PR #43, opened 2020-01-10
Fix script2notebook function

mazurov:
Replace hard coded 'dev/' path with
nbs_path from the settings file


sgugger:
Thanks so much!

PR #44, opened 2020-01-12
CLI to create a new project from template

xnutsive:
Hi there, and thanks for building nbdev! ❤️

I was reading through it's source code notebooks, and figured the best way to learn how nbdev works is to work on it's own codebase. 

This pr adds a new cli tool `nbdev_new` that accepts a name of a new project, and creates the project using the `fastai/nbdev_template` repo as a template. This way, devs cab create new projects in the CLI right after they `pip install nbdev`.

This is a WIP. If this is something you'd like to see in nbdev — I'll be happy to improve it, add some tests, makes sure it works consistently on all platforms.

There are several things I'd like to clean up:

- [x] Make sure this works on Windows
- [x] Write some tests
- [x] Automatically set some variables in `settings.ini`: the project name, github username, and such. 
- [x] Add documentation.
- [ ] Offline support? 

What do you think?


sgugger:
Thanks, this looks useful. Yes to having all the defaults in settings.ini (you can get some inspiration for their values in `Config.create`). Before I merge this, could you build the library? Also let me know if you want to iterate through multiple PRs or polish this one.

xnutsive:
Give me today to check windows support, and tinker with the settings a bit. I'll rebuild the lib, commit everything, and let's merge this, and see what will be the good next step. 

Overall, I'll have some free time this year, so if you'd want help with anything across nbdev, fastai2, fastcore, etc — do let me know what needs work ;)

xnutsive:
@sgugger, I've added `settings.ini` generator that utilizes values from `.gitconfig` if there is one in the user's home directory. 

I've also added unexported cells to test if the command works, and then deletes the folder to clean up. `nbdev_test_nbs` runs those cells fine. 

Haven't tested this on Windows and I suppose paths wouldn't work. 

When I'm done with Windows support, do you want me to also add docs to `index` and `tutorial` notebooks?

sgugger:
Adding docs in  `index` would be great in a separate PR. Not too sure for the tutorial as it wouldn't match the video. Maybe we can add a section at the end "New since the video was made".

Let me know when this is ready for review/merge.

nbdev/cli.py
```diff
@@ -270,4 +271,64 @@ def nbdev_install_git_hooks():
...
+    path = Path(f"./{name}").absolute()
+
+    if path.is_dir():
+        print(f"Directory {path} already exists. Aborting.")
+        return
+
+    print(f"Creating a new nbdev project {name}:")
+
+    try:
+        print("Cloning the template repo")
```
jph00: I'd rather only tell people when things aren't working, rather than telling them about every successful step. Let's remove the print statements except final success

nbdev/cli.py
```diff
@@ -270,4 +271,64 @@ def nbdev_install_git_hooks():
...
+        settings_path = path/"settings.ini"
+        example_settings_path = path/"settings.ini.example"
+
+        if settings_path.is_file():
+            print("Moving an existing settings.ini to settings.ini.example")
+            settings_path.rename(example_settings_path)
+
+        create_template_config(path, name)
+
+        print("Done!")
```
jph00: Maybe say something more descriptive. e.g. "Completed creating repo from template. Please edit settings.ini now, and then then nbdev_build_lib".

sgugger:
Looking good to me, thanks a lot!

PR #47, opened 2020-01-12
Add a warning in case some notebooks have a duplicate Title.

daviddemeij:
Sometimes I copy a Notebook and I forget to change the title, this has caused me some confusion as to why the sidebar apparently was not generated properly. I thought it might be useful to add a simple warning if you are using the same title multiple times when generating the sidebar (as you probably never intend to do this).

Please let me know if you think this is useful, or if I should implement this in another way.


sgugger:
Thanks, this looks useful! Don't forget to run `nbdev_install_git_hooks` before making a PR as the notebook contains metadata that should be stripped to avoid merge conflicts. Could you clean it up before I merge this?

daviddemeij:
Oops I thought I did that, should be good now.

sgugger:
Tanks a lot!

PR #53, opened 2020-01-18
Added nbdev_new paragraph to the index notebook

xnutsive:
I've added a paragraph on `nbdev_new` into the index notebook and ran all cells there.


sgugger:
Thanks! In the future please do not run `make_docs` before submitting your PR and let us do it after it's merge: this makes the PR less readable and easily creates conflicts.

PR #55, opened 2020-01-18
fixed invalid settings bug in nbdev_new

xnutsive:
I got to using my own `nbdev_new` and... it had two bugs in it. 

First, it edited `settings.ini` and `settings.ini.example` _after_ committing the project to the new git repo, and hence after running the command, the user would have one unstaged file (settings.ini) and one file that is not added to the repo at all (settings.ini.example). 

Second, running `nbdev_build_lib` failed, because the settings generated from `create_config` are missing a bunch of attributes (title, descripbion, keywords, etc). 

I'm sorry for introducing the bugs ;( This PR fixes it for now, since it's already in master, but the trade off is that nbdev_new will not generate a config with defaults. 

I want to work on `create_config` so that it would make a config that would actually work, but will also retain all the comments that the default nbdev_template config has — they're valuable and relevant. If you think it's a good idea, I'd work on it in a separate PR.


sgugger:
Noted and thanks! See my comments on the forum for your suggestion.
Also again, please no `make_docs` before submitting the PR. I have fixed the conflicts here but I won't do it next time ;)

PR #56, opened 2020-01-19
Don't give a warning if the index notebook has a duplicate title

daviddemeij:
Building the `fastscript` docs I realized that the `index.ipynb` can have the same title as another notebook (in this case `00_core.ipynb`) and it shouldn't give a warning. This small change should fix that.


sgugger:
Yes, the index being sent to Overview, there is no need. Thanks!

PR #57, opened 2020-01-20
fix nbdev_update_lib when notebooks are in the root folder

maarten990:
With notebooks in the root folder of the repo, the header at the top of the generated Python files looks like this:
```
# AUTOGENERATED! DO NOT EDIT! File to edit: 01_some_notebook.ipynb (unless otherwise specified).
```
but since `Config().nbs_path.relative_to(Config().config_file.parent)` returns `.` in that scenario, `nbdev_update_lib` would look for "./01_some_notebook.ipynb" and the regex search would fail.

This fixes that.


sgugger:
Looks good to me, thanks a lot!

PR #60, opened 2020-01-21
fixes bug that _nb2htmlfname would generate 1.html from 00ab_export_module_1.ipynb

cwza:
I found this bug when I tried to export docs from 2 notebooks named `02_gpt2_lm.ipynb` and `03_lstm_lm.ipynb`.
There is only one html file called `lm.html` be generated from those 2 notebooks.
Therefore I dig into the code in _nb2htmlfname and try to fix it.

The only thing I change is the _re_digits. 
Just add a question mark after greedy match S* and also add a test that check 00ab_export_module_1.ipynb can be transformed to export_module_1.html correctly.

Thanks.


sgugger:
That's perfect, thanks for adding a test that was failing before your fix!

PR #62, opened 2020-01-22
Fix Minor Bug in export2html.convert_nb

hamelsmu:
`exporter` is not a callable

cc: @sgugger


sgugger:
Thanks for the fixes, I went too fast on this yesterday.

PR #63, opened 2020-01-22
Call _nb2htmlfname with dest parameter

hamelsmu:
I noticed that we aren't calling _nb2htmlfname with the dest parameter here, updating for consistency.


sgugger:
Ah yes, forgot it here. Thanks!

PR #65, opened 2020-01-23
raise error instead of print

hamelsmu:
Two changes:

- Raise an error if it occurs, instead of printing the error
- Print the name of the destination file created


sgugger:
No, we don't want the conversion to stop if there is an error (which raising an exception will do). We still want the other notebooks to be converted.

hamelsmu:
@sgugger shall I make an option available to raise an error, with the default being to print as it was before - or just revert that part of it completely?

sgugger:
Not sure what's your use case, as errors don't play nicely with parallelism. Do you use only one worker with this?

sgugger:
In that case, I would use the same thing we do in `nbdev_test`: return True if no error and False if there is an error (on top of printing it) in `_notebook2html`, then raise an exception in `notebook2html` if there was one error at least, after the whole conversion has been done.

I can do this in a couple of hours.

sgugger:
Did it in [this commit](https://github.com/fastai/nbdev/commit/54e0723511ee423f5a8aa69722502e002998bb47) so closing this PR.

PR #66, opened 2020-01-23
Add CI Badge to README

hamelsmu:


sgugger:
Oh I merged this, but it 's going to be deleted the next time we make the documentation, so this won't really work. It should be added to the index notebook.

sgugger:
It ended up being harder than I thought, so I added a PRE_README that can be edited by the user as they want (and is added at the beginning of the README, before the result of the index notebook conversion).

PR #68, opened 2020-01-24
Set min nbconvert version due to API change

evan-burke:
Fixes an issue with older versions of nbconvert - looks like prior to 5.6.1 or so - that have 'resources' as a required argument to the preprocess function, resulting in errors like this when running tests:
 `preprocess() missing 1 required positional argument: 'resources'`

Is this the only place the version would need to be updated?

nbconvert commit for the API change:
https://github.com/jupyter/nbconvert/commit/ad6d34b563d1a137a19173c149b85fc11876e2d8

### DIFF
--- a/settings.ini
+++ b/settings.ini
@@ -14,7 +14,7 @@ status = 2
 min_python = 3.6
 audience = Developers
 language = English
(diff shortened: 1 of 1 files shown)


sgugger:
Thanks a lot! Yes this is the only place it's needed.

PR #69, opened 2020-01-27
correct else statement

hamelsmu:
I encountered an error in my testing

> UnboundLocalError: local variable 'passed' referenced before assignment

This was because previously if no notebooks changed, then the variable `passed` would not have been created


sgugger:
Oh yes, bad indent. Thanks a lot for fixing!

PR #73, opened 2020-01-30
Fix minor error in else statement

hamelsmu:
The result of `glob.glob(fname)` is a List[str], which caused an error because later on expects List[Path]

Learned about this error when trying to reproduce https://github.com/fastai/fast_template/issues/18

Without this fix will get this error:

```
AttributeError: 'str' object has no attribute 'name'
```


sgugger:
Looking good to me, thanks for the fix!

PR #75, opened 2020-01-30
escape `{{` curly braces that might be in generated HTML from notebook

hamelsmu:
escape `{{` curly braces that might be in HTML

I've encountered this while using this template, and this helps to guard against this situation where it may cause a problem with liquid


sgugger:
Thanks!

PR #76, opened 2020-01-30
Fix Bug in #73

hamelsmu:
@sgugger  I introduced a bug in https://github.com/fastai/nbdev/pull/73 🤦‍♂ .  I am really sorry about this.  To try to mitigate this from happening I added tests for `notebook2html`.  I am not 100% sure this is the way you want to see tests, but I tried to follow the conventions in the notebook carefully.  

Happy to amend this PR with any suggestions.  🙇  Thank you


nbdev/export2html.py
```diff
@@ -391,7 +391,9 @@ def notebook2html(fname=None, force_all=False, n_workers=None, cls=HTMLExporter,
     "Convert all notebooks matching `fname` to html files"
     if fname is None:
         files = [f for f in Config().nbs_path.glob('*.ipynb') if not f.name.startswith('_')]
-    else: files = list(p.parent.glob(p.name))
+    else:
+        p = Path(fname)
+        files = list(p.parent.glob(p.name))
```
hamelsmu: This was the bug.  I forgot to include `p = Path(fname)`.  I realized that I didn't catch this error because I didn't have great tests.  So I added some tests to this notebook.

hamelsmu:
I also turned on CI in my local repo and can verify [the latest commit it is passing](https://github.com/hamelsmu/nbdev/runs/418039359?check_suite_focus=true)

image

sgugger:
Thanks! I guess I should also have read more carefully ;)
The test seem fine, I may just refactor them using fastcore one of these days.

sgugger:
Follow-up question. Do you know if there is a way to automatically trigger the CI at each commit or at the least with an action form me before merging (like adding a special label)? That would help ensure these kinds of breaks don't happen.

hamelsmu:
I’m looking into this now, will comment back on this thread when I find the best solution

hamelsmu:
Opened https://github.com/fastai/nbdev/pull/77 to address this

PR #77, opened 2020-01-31
Enable CI on Pull Requests

hamelsmu:
@sgugger this will help solve the problem to catch things like what happened in #73

You will also have to turn on branch protection.   In settings, create a branch protection rule for the `master` branch and check `Require Status Checks To Pass Before Merging`

image

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,5 +1,5 @@
 name: CI
-on: [push]
+on: [push, pull_request]
 jobs:
   build:
     runs-on: ubuntu-latest


sgugger:
Perfect! I don't want to necessarily protect the branch (as I sometimes want to merge things that fail and fix them myself) but this is great. Could you update the template as well? I'm going to push this to our other nbdev-generated repos.

hamelsmu:
Sorry What do you mean by update the template?  fast_template?

sgugger:
nbdev_template (thanks again for all those PRs :) )

PR #82, opened 2020-02-09
Updated file exists assertion message to point to the correct function.

hnrkptrsn:
__Config.\_\_init\_\_ assert message updated__
The message on the failed assert for existence of settings.ini refers to a non-existing Config.create method. In the notebooks create_config is used, and I have update the assert message accordingly.


sgugger:
Thanks!

PR #83, opened 2020-02-11
fix nbdev_new when path including space

cwza:
Bug happened when I try to use nbdev_new in the directory including space like "/gdrive/My Drive/xxx".
Fix it by not using split() at target path.


sgugger:
Thanks!

PR #86, opened 2020-02-12
remove raw tags as they break `fastpages`

hamelsmu:
There is liquid injected into fastpages notebook markdown to allow for:

- jekyll notes
- embedded youtube videos
- embedded twitter cards

So will have to find another way to escape `{{`  in notebooks with a special template that escapes code blocks only.  I will make a PR for this soon.

If you do happen to merge, this if you could please update pypi 🙇


hamelsmu:
Nevermind.  I think for right now it makes sense for fastpages to maintain its own template, as I can see both projects having further collisions.  

I'll make a pull request in the future with a more targeted change.

PR #87, opened 2020-02-12
Adding Collapsable Code Cells

hamelsmu:
@inc0 and I discussed this with @jph00 over email, and one way to achieve this is using `<detail>` blocks.  Summary of changes:

- Updated `export2html` to add cell comment shortcut `#collapse` and `#collapse-show` depending on whether or not you want the collapsible element to be open or closed by default.  
- Updated the template to wrap cells in `detail` tags with appropriate css classes.
- Added CSS that dynamically changes the collapsable heading text from `Show Code` to `Hide Code` depending upon if the element is open or closed.

Please let us know if you would prefer we take a different approach.  Thank you.


hamelsmu:
P.S. I noticed tests failing on the master branch prior to this PR, it is not obvious to me why tests are failing but I don't believe it is something in this PR.

jph00:
Looks like a great approach!

hamelsmu:
@jph00 / @sgugger Any chance someone can cut a release to pypi so I can incorporate this into `fastpages` ?  Thank you 🙇

jph00:
Done.

hamelsmu:
@jph00 / @sgugger Wohoo!  We have collapsible code cells now in `fastpages`

collapse-cells

https://fastai.github.io/fastpages/2020/02/12/test.html

PR #90, opened 2020-02-13
fix create_mod_file to replace \ for /

tcapelle:
The windows file sep is `\` and the linux one is `/`, so the header changes each time the lib is built on windows/linux. Solves the issue by replacing the `\` by `/` on the header.
Avoids this:
`# AUTOGENERATED! DO NOT EDIT! File to edit: nbs\01_core.ipynb (unless otherwise specified).
`


tcapelle:
finally I managed to do it cleanly ufff....

sgugger:
Seems reasonable, thanks a lot!

PR #91, opened 2020-02-13
add collapse alias with hyphen

hamelsmu:
I found people were making mistakes in adding this option by often using a hyphen instead of underscore, so adding this minor change

### DIFF
--- a/nbdev/export2html.py
+++ b/nbdev/export2html.py
@@ -177,10 +177,10 @@ _re_cell_to_remove = re.compile(r'^\s*#\s*(hide|default_exp|default_cls_lvl)\s+'
 
 # Cell
 #Matches any cell with #collapse or #collapse_hide
-_re_cell_to_collapse_closed = re.compile(r'^\s*#\s*(collapse|collapse_hide)\s+')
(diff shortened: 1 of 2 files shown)


sgugger:
One other thing you could consider is checking the cell metada for 'collapsed: True' (if cell output is toggled with o, this is added).

hamelsmu:
> One other thing you could consider is checking the cell metada for 'collapsed: True' (if cell output is toggled with o, this is added).

Thanks, Sylvian, just so I understand are you thinking about the Collapse Heading or Hide Cells Jupyter plugin?  I know the hide cells plugin is already used by nbdev.

The Collapse Heading is an interesting idea.  Any suggestions on the smoothest user interface to enable/disable this kind of thing?  I have been thinking about it haven't come up with a good idea yet.  

Please let me know your thoughts.  Thanks

sgugger:
I think it comes with collapse headings. Like I said, it adds some metadata in the cell, so if you check for it you can have collapsed outputs be treated directly.

hamelsmu:
Okay, thanks I will look into this soon

PR #97, opened 2020-02-24
Add a new Keyword "# exporti" and minor tweaks

flpeters:
Hi,
I've recently started converting a project to use nbdev, and found that I needed the ability to export a cell to the .py file, but not include it's content in `__all__`, or the documentation.

That's why I've added an additional keyword, called `exporti`. It stands for "export internal", and can be used instead of `#export` or `#exports`. 

Below is a __change log__ of everything that's changed. __Only two files have been changed__, namely `nbs/00_export.ipynb` and `nbs/03_export2html.ipynb`. The other files in this commit were generated.

I hope this feature is along the lines of what you intend nbdev to be.

The __short version__ is:
for exporting to lib:
A new check has been added to `is_export()` and the result of that check is used to decide whether or not the content of a cell should be added to `__all__` in `_notebook2script()`.
and for exporting to docs:
The new Keywords has been added to `_re_cell_to_remove`.
Plus a small change I made while working to understand the code (which doesn't have anything to do with this feature, but I left it in because it made it so much easier to understand the code).

The __long version__ is:

nbdev_export_lib `nbs/00_export.ipynb`:

- changed `_re_blank_export` to also match any `#export` with an "i" at the end, like: `#exporti`. all other previous rules also still apply.

- changed `_re_mod_export` to also match any #export with an "i" at the end, like `#exporti bla`. Also, it is now required for there to be at least one whitespace after the `#export[is]?`, before any sub-modules are listed. This is in my opinion the correct behaviour, because e.g. `exportsome.module` is syntactically ambiguous due to the export[s]. previously that example would have resolved to `exports ome.module`, which is not what someone might expect.

- added `_re_internal_export`, which is used to detect if `#exporti` is used anywhere in a cell.

- changed `is_export()` to now also check if the export statement `#exporti` is used, by using the new `_re_internal_export` mentioned above. `is_export()` now returns a tuple, with the previous return value as it's first element and either `True` or `False`, depending on if `#exporti` was used or not. The boolean is `True`, if `#exporti` was __not__ used, and False if `#exporti` was used. The value represents whether or not the code should be added to `__all__`. If no match was found then the function returns `None`, like before.

- changed `_notebook2script()` to use the new `is_export()` correctly. The tuple is unpacked after `None` values are filtered out, and the new boolean value is used to determine whether to add the content of the code block to `__all__`.

- changed `_notebook2script()` to put a comment saying `# Internal Cell` in the output code, instead of the previously used `# Cell`, in cases where `#exporti` is used.

- changed the `is_export()` tests to reflect the changes to the return type.

- added new tests for `is_export()`, to test the new `#exporti` keyword.

- added new tests for `is_export()`, to test the changed semantics for including a sub-module.

- changed an existing `is_export()` test that tested specifying a different sub-module, to now use the os specific path separator, after it failing on my windows machine.
The previous expected output was `mod/file`. When I ran the test, the output was `mod\file` or rather `mod\\file`.

- changed the documentation at the very top if the file to include a description of the new keyword.

- changed the documentation for `is_export()` to include the new keyword.

- changed `_re_default_exp` to, similar to `_re_mod_export`, now require at least one whitespace character after the `default_exp` to be a valid expression. The reason is that same as above, namely that something like maybe `#default_experson` would become `#default_exp erson`. Which is a silent bug waiting to happen and difficult to track down because nbdev doesn't notify you. In my opinion it's better to strictly require the user to put a whitespace between the keyword and the argument.

- changed documentation of `_re_default_exp`, which included a copy-paste artifact.

nbdev_export_docs  `nbs/03_export2html.ipynb`:

- changed `_re_cell_to_remove` to also match on cells containing the new `#exporti` keyword.

- changed the documentation of `remove_hidden()` to reflect the change above.

- added a test case for this new keyword to the `remove_hidden()` test.

- changed `convert_nb()` by pulling function composing out of the for loop. This is a change I made while trying to understand how the code worked and it seems better this way, so I left it in.

__Issues__:
- Probably due to os specific file I/O behavior, operations that overwrite existing files fail on my windows 10 machine. That's why the cell output of one cell in 03_export2html has changed.
- Tests that use hard coded path separators fail on my windows machine, probably because windows uses "\\" and linux uses "/".


sgugger:
Wow, what a nice PR! I was too lazy to add this since you can always underscore private functions but this is looking really good. Thanks a lot!

PR #98, opened 2020-02-25
Improve support for pythons type hint syntax (pep 484)

flpeters:
Hi again,
I'm working a lot with ctypes at the moment, which is why it's much easier in the long run to write code where most things have type annotations on them.
I've noticed however, that when I declare a global variable with a type hint e.g. `a:int = 1`, the name that's added to `__all__` is not `'a'` as one might expect, but `'a:int'`.
This causes any code that is trying to import that module to crash, because `'a:int'` is not valid python syntax for a variable name.

This PR is fixing this particular case, and a few other things I spotted along the way.

I hope this will be useful to people.

Here is a __changelog__ of everything that's changed. All changes were made in `nbs/00_export.ipynb`:

- Add a test case to `_re_patch_func` to check that type annotations, including return types, are handled correctly.

- Change `_re_patch_func` to require that there is at least one character used as a function name. Previously `def (...)` would also match.

- Delete repeated and identical declaration of `_re_patch_func`.

- Add support for return type annotations to `_re_typedispatch_func`. Previously a function that had a return type specified using the arrow notation `def ... (...) -> type:` would not match.

- Change `_re_typedispatch_func` to require there to be at least one character used as a name. Previously `def (...)` would also match.

- Changed `_re_typedispatch_func` documentation to reflect the changes above, and correct a copy-paste artifact.

- Add a test case to `_re_typedispatch_func` to check that return types are handled correctly.

- Change `_re_class_func_def` to require there to be at least one character used as a name. Previously `def (...)` would also match.

- Add test case to `_re_class_func_def` to check that type annotations are handled correctly.

- Change `_re_obj_def` to correctly distinguish type information from the variable name. Previously `a:int = 1` would match as `a:int`, which is not correct. Also variable names will now only match if they are a legal python name. Previously e.g. `@-@./! = 1` would match as `@-@./!` and `@dec(a=1)` would match as `@dec(a`, which could be a problem when dealing with decorators (although the decorator case is caught by a later check, this is in my opinion still the cleaner way).

- Change documentation of `_re_obj_def` to make clear how this new code works.

- Add test cases to `_re_obj_def` that check for some of the edge cases mentioned above.

For more info on pythons type hints view https://www.python.org/dev/peps/pep-0484/

### DIFF
--- a/nbdev/export.py
+++ b/nbdev/export.py
@@ -101,7 +101,7 @@ _re_patch_func = re.compile(r"""
 @patch         # At any place in the cell, something that begins with @patch
 \s*def         # Any number of whitespace (including a new line probably) followed by def
 \s+            # One whitespace or more
-([^\(\s]*)     # Catch a group composed of anything but whitespace or an opening parenthesis (name of the function)
+([^\(\s]+)     # Catch a group composed of anything but whitespace or an opening parenthesis (name of the function)
 \s*\(          # Any number of whitespace followed by an opening parenthesis
 [^:]*          # Any number of character different of : (the name of the first arg that is type-annotated)
 :\s*           # A column followed by any number of whitespace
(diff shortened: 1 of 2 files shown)


sgugger:
That all seems perfectly reasonable. Thanks for adding all those tests when fixing functionality!

PR #101, opened 2020-03-01
remove anchor tag text

hamelsmu:
# Background

- nbconvert injects the ugly ¶ symbol as text in anchor tags.
- nbdev ignores the ¶ using CSS by [setting display value to none](https://github.com/fastai/nbdev/blob/c41b7e6b85ae452fb1d9eeb5fe7077275eb0c1ef/docs/css/customstyles.css#L1-L3) and uses anchor.js to apply beautiful styling. 
- However, these ugly ¶ show up in fastpages, and I'm using javascript to remove them.  This seems like the best way to remove this text.   It is useful to remove this text b/c it shows up in search results.

cc: @sgugger


sgugger:
Thanks for fixing this!

PR #102, opened 2020-03-03
Fix #101

hoftherose:
Fix what seems to be breakage after #101, just adding " " (space) instead of empty string to exporter.anchor_link_text. 

With an empty string the html generates self closing "a" tags. When inside of a header (which happens in every header with nbdev) it expects a following opening tag but instead gets a closing one. This makes it so that you can't see text until you add a new link, whenever that may be. Don't quite understand why this happens though. 

Html seems to ignore the extra whiteline so you can't see it, it doens't even indent text.

Images of before and after: 

Screenshot_20200302_195154 

After

Screenshot_20200302_195417

### DIFF
--- a/nbdev/export2html.py
+++ b/nbdev/export2html.py
@@ -364,7 +364,7 @@ def nbdev_exporter(cls=HTMLExporter, template_file=None):
     exporter = cls(cfg)
     exporter.exclude_input_prompt=True
     exporter.exclude_output_prompt=True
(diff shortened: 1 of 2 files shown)


hamelsmu:
Wow 😳 really sorry about this.  I should render the full nbdev docs and test that next time

sgugger:
Thanks a lot for fixing!

Issue #106, opened 2020-03-19
Inline images aren't being displayed on nbdev.fast.ai


danielpcox:
I notice that some images aren't being correctly displayed, at least on the front page of nbdev.fast.ai:

![image](https://user-images.githubusercontent.com/56411/77087502-96f9e280-69d9-11ea-9ef1-7fc6b9f61de7.png)

sgugger:
I have no idea why it's now happening. @hamelsmu so you think there is anything we did for fastpages that could have changed how those {% include image.html ...} work?

hamelsmu:
What page are we seeing this on?  

![image](https://user-images.githubusercontent.com/1483922/77203676-35577800-6aae-11ea-8dc6-abad51fd57ba.png)

Is this the sidebar?  That shows up for me ... is it something else that isn't rendering?

sgugger:
No, not the sidebar, the block with {% include image.html ...} should be rendered as an image (and it used to work).

hamelsmu:
Oh I see it now!  Yes this probably has something to do with escaping liquid with raw tags.  Taking a look

hamelsmu:
I opened two separate PRs to fix this issue.  Thanks

danielpcox:
Thanks guys!

PR #108, opened 2020-03-20
Update jekyll.tpl

hamelsmu:
This closes https://github.com/fastai/nbdev/issues/106

### DIFF
--- a/nbdev/templates/jekyll.tpl
+++ b/nbdev/templates/jekyll.tpl
@@ -11,8 +11,6 @@ sidebar: home_sidebar
 {% include 'autogen.tpl' %}
 
 <div class="container" id="notebook-container">
-    {{ "{% raw %}" }}
         {{ super()  }}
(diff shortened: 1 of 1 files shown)


hamelsmu:
@sgugger I can open a separate PR, the original purpose of these raw tags were to escape liquid in notebooks which were causing things to break, especially in code cells.  

I can make a separate PR if you wish to target code cells only, this fixes the problem for now

sgugger:
Escaping them in code cells sounds fine to me, if you don't mind doing a separate PR. Ideally it would be nice to know what things are broken so we escape those specifically.

hamelsmu:
The things that are broken is when people try to put `{{` in their code, which for example is common if someone is using Keras  callbacks to format stuff.  I don't think there is any way around it except to escape the code block for liquid

sgugger:
Ok, so the fix sounds right then.

PR #109, opened 2020-03-20
Only escape Liquid in Code Blocks

hamelsmu:
This is related to https://github.com/fastai/nbdev/pull/108


sgugger:
Thanks a lot for the swift PRs!

PR #112, opened 2020-03-25
Change Linux specific check to also work on other platforms

flpeters:
On windows the root of the file tree is a drive letter (e.g. `'C:/'`), meaning the `cfg_path != Path('/')` check can never be false, causing an infinite loop when no settings.ini exists.


sgugger:
Thanks for fixing!

PR #118, opened 2020-03-31
Add #hide-output and #hide_output flags

hamelsmu:
This closes https://github.com/fastai/nbdev_template/issues/12#event-3184124678

EDIT: I guess I was solving the wrong problem and the original issue was about having collapsible output?  But I couldn't find the non-collapsalble functionality in nbdev even though it looks like it was supposed to be here?  I might be confused.

cc: @sgugger


hamelsmu:
closing this PR for a moment, will reopen

PR #122, opened 2020-04-03
Gitlab Support

tcapelle:
This PR adds basic support for GitLab, solving 2 things:
- Adds backlink to gitlab on the top navigation for the documentation.
- Add host flag on `settings.ini` file, this can be used later by `Config`

>It does not solve yet the source links correctly to point back to the source code.


tcapelle:
You will need to add the `host: github` to the `nbdev_template`

sgugger:
Thanks for the PR!
There is one problem left with backward compatibility. In imports.py, you should also add `host` to the `_defaults` dict, so that for all current users who don't have `host` in their config file, it's automatically set to github and their projects are not broken. Could you adjust that?
Then we'll need a follow-up PR in nbdev_template.

tcapelle:
Here you have the `nbdev_template`: [PR](https://github.com/fastai/nbdev_template/pull/13/files)

sgugger:
Thanks a lot for all your work on this!

PR #124, opened 2020-04-05
fix: allow use of other decorators below patch

borisdayma:
I believe `__all__` and `_nbdev.py` were not correct due to extra decorators being between `@patch` and the function definition.
This should fix it.
Another test has been added.


sgugger:
Thanks a lot for this!

PR #126, opened 2020-04-05
fixes #123

### LINKED ISSUE #123: Hyperlinks in documentation don't work with ">" blocks
Related with #107 , I updated to the latest version `0.2.17` and I'm getting the same error.

[Here](https://lgvaz.github.io/fastcook/) is an example.

This problem does not happen at the exported `README.md` file

---
**Note:**
Can I be running on a old version? I checked my version by doing:
```python
>>> import nbdev
>>> nbdev.__version__
'0.2.17'
```
But does this mean that this is the version used when I run `nbdev_build_docs`?
I think I'm running the right version, because I installed everything under the same conda virtual env, but I'm leaving this note because the problem was fixed in #107 and I'm still facing the issue, so I can be doing something wrong here xD

lgvaz:
First attempt at fixing #123

### DIFF
--- a/nbdev/export2html.py
+++ b/nbdev/export2html.py
@@ -313,6 +313,16 @@ _re_properties = re.compile(r"""
 (.*?)$     # Any pattern (shortest possible) then end of line
 """, re.MULTILINE | re.VERBOSE)
 
+_re_mdlinks = re.compile(r"\[(.+)]\((.+)\)", re.MULTILINE)
+
+# Cell
+def _md2html_links(s):
+    matches = _re_mdlinks.findall(s)
+    for name,ref in matches:
+        s = s.replace(f'({ref})', f'{name}</a>')
+        s = s.replace(f'[{name}]', f"<a href='{ref}'>")
+    return s
+
 # Cell
 def get_metadata(cells):
     "Find the cell with title and summary in `cells`."
@@ -323,8 +333,8 @@ def get_metadata(cells):
                 cells.pop(i)
                 attrs = {k:v for k,v in _re_properties.findall(cell['source'])}
                 return {'keywords': 'fastai',
-                        'summary' : match.groups()[1],
-                        'title'   : match.groups()[0],
+                        'summary' : _md2html_links(match.groups()[1]),
+                        'title'   : _md2html_links(match.groups()[0]),
                         **attrs}
 
     return {'keywords': 'fastai',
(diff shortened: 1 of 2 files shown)


lgvaz:
Do I need to be more careful when looking for markdown links? I first started with the regex described [here](https://davidwells.io/snippets/regex-match-markdown-links) but I found it to be too restrictive.

nbdev/export2html.py
```diff
@@ -323,8 +334,8 @@ def get_metadata(cells):
                 cells.pop(i)
                 attrs = {k:v for k,v in _re_properties.findall(cell['source'])}
                 return {'keywords': 'fastai',
-                        'summary' : match.groups()[1],
-                        'title'   : match.groups()[0],
+                        'summary' : _md2html_links(match.groups()[1]),
+                        'title'   : _md2html_links(match.groups()[0]),
```
sgugger, line 338: Titles should be free of links so I'm not sure we need one here.

nbdev/export2html.py
```diff
@@ -313,6 +313,17 @@ def remove_empty(cells):
 (.*?)$     # Any pattern (shortest possible) then end of line
 """, re.MULTILINE | re.VERBOSE)
 
+_re_mdlinks = re.compile(r"\[(.+)]\((.+)\)", re.MULTILINE)
+
+# Cell
+def _md2html_links(s):
+    'Converts markdown links to html links'
+    matches = _re_mdlinks.findall(s)
+    for name,ref in matches:
```
sgugger, line 322: I think this might be faster and cleaner to implement this as a replace? I haven't tested, so it might need to be adjusted but I think just returning
```
_re_mdlinks.sub(r'<a href='\1'>\0</a>, s)
```
should work.

lgvaz:
I'm having some problems with my git_hooks because I'm jumping between computers. Give me a moment to clean this up =)

nbdev/export2html.py
lgvaz: Brilliant!!

lgvaz:
Alright, after a thousand commits I got it! lol

lgvaz:
The `re.sub` still blows my mind, I tried to use it before the initial solution, but I was getting confused on how the parameter groups would work, and how the `[]()` would be handled after the substitution.

But everything just works, simple as that. I really like this solution haha

nbdev/export2html.py
lgvaz: Done!

sgugger:
Thanks a lot!

PR #130, opened 2020-04-11
Local Jekyll rendering incorrectly with extra `/`

noklam:
Removing ending "/" for correct URL
 #Related Issue
#126 Local Jekyll rendering incorrectly with extra `/`

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -57,6 +57,6 @@ defaults:
 
 sidebars: [home_sidebar]
 permalink: pretty
-baseurl: /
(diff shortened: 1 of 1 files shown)


sgugger:
I am unsure this will not break the actual docs, so may revert this if it does.

sgugger:
Looks ok, thanks!

noklam:
Thanks for accepting the PR. I should have mentioned at first that I have tested it with the original docs.

Thanks @sgugger,  love nbdev!

PR #133, opened 2020-04-16
build_docs does not overwrite top bar config files

kikejimenez:
Hi,
After built with `nbdev_build_docs`,  for a given settings.ini file, I modify the *user* (in the settings.ini)
 and I build the docs again with `nbdev_build_docs`. The top bar of the documentation keeps pointing to the first *user*  home and GitHub URLs.

This simple solution has worked for me:
I removed the conditional that prevents the overwriting of the configuration files.

Thanks for this great package!


sgugger:
No, we don't want to overwrite the templates since they are default provided to users, and users may want to change them. If we take your change, those customizations would be erased at each new build.

kikejimenez:
Oh!, I see your point, thanks!

PR #135, opened 2020-04-17
_script2notebook: assertion wrongly reports '0' for expected number of cells

tchklovski:
in
`assert len(splits)==len(dic[key]), f\"Exported file from notebooks should have {len(dic[fname])} cells but has {len(splits)}.\"`
what's checked is `len(dic[key])` but what's reported in assertion failure message is `len(dic[fname])`. since `dic` is
a defaultdict, it's just happy to output the incorrect `0`


sgugger:
Thanks!

PR #136, opened 2020-04-23
fixed nbdev_update_lib to work with exporti

pete88b:
Using #exporti causes nbdev_update_lib to fail - I added another test only cell in 00_export to demo.
Without the "^# Internal Cell" part of _re_cell, nbdev_update_lib fails with: 
  "AssertionError: Exported file from notebooks should have 37 cells but has 36."
Hope you don't mind but I also updated the error message to tell us which file failed to update. new message would read: 
  "AssertionError: "export.py" exported from notebooks should have 37 cells but has 36."


sgugger:
LGTM, thanks!

PR #139, opened 2020-04-24
add multi-line for flags

hamelsmu:
per https://forums.fast.ai/t/fastpages-github-pages-blog-using-nbdev/62828/231

- adds alias for hide_input to include hide-input
- adds multiline detection of flags
- makes flags case insensitive 

cc: @sgugger


sgugger:
LGTM, thanks a lot!

Issue #137, opened 2020-04-23
Does it matter that _all_ can't have line breaks?


pete88b:
If we use line breaks between the elements in _all_, they don't get added to __all__
See: https://github.com/pete88b/decision_tree/blob/master/60_all_test.ipynb
If this is intentional, I can update the docs - if not, I'd be happy to change the regex and test etc

while we're in that area: should _add2add be called _add2all?

sgugger:
Yes `_add2add` should definitely be named `add2all` if you want to make that PR. 
As for the line break, tt's not intentional, it just never came up. If it's not too complicated to add, happy to support it, otherwise let's document it and encourage people to do separate `_all_` on separate cells.

pete88b:
Nice one Sylvain; Please see https://github.com/fastai/nbdev/pull/140

I've left _all_ without multi-line support as this fits better with an idea i've had for an %nbdev_add2__all__ magic flag ...

sgugger:
Ok, Closing this then and let's see what your idea is :)

PR #140, opened 2020-04-26
add2add->add2all and updated docs for extra_add

pete88b:
wrt: https://github.com/fastai/nbdev/issues/137
- renamed _add2add to _add2all
- updated docs to include detailed use of `_all_`
- hidden the _add2all test cell - not sure it makes sense to see tests for a function that is not in the docs


sgugger:
Thanks! In general though, please don't build the docs when making a PR as it makes the diff hard to read. We prefer to build the docs in separate PRs.

PR #142, opened 2020-04-28
make black and make sync

zonca:
* `make black`: run black on the exported library
* `make sync`: synchronize changes in the exported library back to the notebooks


sgugger:
I have no problem with make sync which is linked to `nbdev`. I wouldn't put `make black` in the template though, as it's not directly linked to nbdev (people can in any case change their CMake file as they wish once it's generated).

zonca:
ok

sgugger:
Thanks!

PR #144, opened 2020-04-29
Installation and Tutorial clarifications

### LINKED ISSUE #141: Undocumented assumption in "installation" and "tutorial"
_Disclaimer: I'm a conda/jupyter n00b user - I've always run virtualenv specific jupyter servers. Which is a hassle with the changing ports. So I may be missing something "obvious" about how these work together._

The instructions say "`pip install nbdev`", however they don't say in what context it should be done. The video gives the impression conda/jupyter/nbdev can be setup so your default jupyter invocation points to a parent directory of where the `nbdev` template repository is cloned.

While any help is appreciated, the ideal layout (for me) would be:
a) a conda managed jupyter environment
b) a choice of jupyter kernels for different python versions (3.x) (i.e. for generic notebook creation)
c) separate conda environments per project

Here are the questions I have after reading & listening to the tutorial:
- Which of those environments need to have `nbdev` installed into them? 
- When does the environment need to be activated? (There's no mention in the tutorial.)
- What is your workflow to allow a single jupyter deployment to service multiple repositories?

hwine:
Be clearer that `nbdev` needs to be installed in the same environment as
used by the Jupyter Server and any terminal sessions.

Fixes #141

### DIFF
--- a/nbs/index.ipynb
+++ b/nbs/index.ipynb
@@ -81,7 +81,9 @@
     "```\n",
     "git clone https://github.com/fastai/nbdev\n",
     "pip install -e nbdev\n",
-    "```"
+    "```\n",
+    "\n",
+    "_Note that `nbdev` must be installed into the same python environment that you use for both your Jupyter Server and your workspace._"
    ]
   },
   {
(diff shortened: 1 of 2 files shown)


sgugger:
Thanks a lot!

Issue #149, opened 2020-05-07
settings.ini modified in github actions


bepuca:
I am working on a project using nbdev. It worked flawlessy so far but now the following problem is happening:

## Observed behavior
- When pushing a certain branch, github actions are triggered properly but breaks on `Check if all notebooks are cleaned` with the error message `git status is not clean`. Now, I added some more logging to try and see what exactly is not clean.
- It seems that the file `settings.ini` is modified when installing the library, leading to non-clean git status.
- Here are the logs on what exactly is changed:
```
diff --git a/settings.ini b/settings.ini
index
--- a/settings.ini
+++ b/settings.ini
@@ -1,5 +1,4 @@
 [DEFAULT]
-# All sections below are required unless otherwise specified
 lib_name = my_lib_name
 user = fastai
 description = my_description
@@ -12,42 +11,16 @@ version = 0.0.1
 min_python = 3.6
 audience = Developers
 language = English
-# Set to True if you want to create a more fancy sidebar.json than the default
 custom_sidebar = False
-# Add licenses and see current list in `setup.py`
 license = apache2
-# From 1-7: Planning Pre-Alpha Alpha Beta Production Mature Inactive
 status = 2
-
-# Optional. Same format as setuptools requirements
 requirements = numpy pandas matplotlib requests tqdm pyarrow joblib fastai torch xlrd hyperopt scikit-optimize
-# Optional. Same format as setuptools console_scripts
-# console_scripts = 
-
-###
-# You probably won't need to change anything under here,
-#   unless you have some special requirements
-###
-
-# Change to, e.g. "nbs", to put your notebooks in nbs dir instead of repo root
 nbs_path = nbs
 doc_path = docs
+doc_host = https://fastai.github.io
+doc_baseurl = /my_lib_name/
+git_url = https://github.com/fastai/my_lib_name/tree/master/
+lib_path = my_lib_name
+title = my_lib_name
+host = github
 
-# Anything shown as '%(...)s' is substituted with that setting automatically
-doc_host =  https://%(user)s.github.io
-doc_baseurl = /%(lib_name)s/
-git_url = https://github.com/%(user)s/%(lib_name)s/tree/%(branch)s/
-lib_path = %(lib_name)s
-title = %(lib_name)s
-
-#Optional advanced parameters
-#Monospace docstings: adds <pre> tags around the doc strings, preserving newlines/indentation.
-#monospace_docstrings = False
-#Test flags: introduce here the test flags you want to use separated by |
[...]

PR #150, opened 2020-05-07
helpful msg for git status error

bepuca:
Following the problem reported in #149 , add a bit more information on the error message when git status is not clean so users facing the same problem are provided with a course of action to try and fix it.


sgugger:
Thanks!

PR #152, opened 2020-05-13
Magic flags

pete88b:
Change nbdev to use magic flags instead of special comments - see: https://forums.fast.ai/t/can-we-have-tab-completion-and-help-for-nbdev-flags/64603
Please let me know if you'd like a call to walk through the changes - there are a lot of changes and I still have a few questions (o: and we'll need to co-ordinate changes to github/gitlab templates etc
I'm also hoping that this is the first of a few rounds of changes and we can follow-up with more ideas like using magic flags for _all_ (but I didn't want to put too much into this PR)


sgugger:
This is a very consequential change that is not easy to implement now that nbdev is used as a fundation for fastpages. We can't break backward compatibility and expect users to change their habits in a day.

At the same time, autocomplete with magic is a nice thing to explore, as is the help-display it can enable. So I don't want to throw away the idea. I think the right direction would be to support *both* comments and magic commands, to save backward-compatibility and enable this cool feature. It would also make the migration easier.

Fair warning: I have almost no time to work on this right now and limited time to review. But if you're up to it, we could merge this more progressively: just add a %nbdev_export first and make everything work with #export or %nbdev_export, check this all works well and progressively add the new flags.

(Sidenote: please don't build the docs when you submit a PR as it makes the diff unreadable. A follow-up PR that just updates the doc (or I can push that change easily) is preferable.)

sgugger:
cc @hamelsmu

hamelsmu:
I generally like the magic command approach and I see why it is appealing.  

One thing that might give everyone some comfort is fastpages versions are pinned to specific versions of nbdev.  So if this is eventually merged, it will not break fastpages **unless people upgrade to a new version**.   In that case, I suppose we could make a conversion script that automatically executes upon an upgrade ( we can trigger this automatically if it is detected that someone is upgrading from version < 2.x to version > 3.x).  That way we could mitigate the effect of any breaking changes.  

TLDR; If we want to continue working on this, I think a conversion script to convert all notebooks to use magic commands is necessary to automatically mitigate breaking changes.  

Let me know if anyone has any thoughts about this, in theory, a conversion script doesn't seem like it would be that complicated.

pete88b:
Think it makes sense to close this PR as we're adding magic flags progressively

PR #153, opened 2020-05-15
Adding %nbdev_export

pete88b:
couple of notes that might help review;
- nbdev/sync.py has changed because we have to `split_flags_and_code` on export - so that we don't write magics to the lib
- copied 00_export.ipynb and 07_clean.ipynb to test folder - so we have some "comment flag" nbs to test with
- I'm testing with fastai2 and have set-up [a test project](https://github.com/pete88b/decision_tree) to cover as many edge cases as I can think of


sgugger:
This all looks good to me though I'll need to test it on some projects before merging. Before I do that, could you update all your code to the convetion: "1-line doc comment in double quotes" for all functions and remaining of the docs in a markdown cell in the notebook after the function definition?
In particular, there is no need to replace current doc comments (unless it was deadly wrong) but you can add more info in the notebook.
Thanks!

pete88b:
I've removed all multi-line comments except for 
```python
def nbdev_export(line):
    """Put an `%nbdev_export` magic on each cell you want exported but not shown in the docs.
    Optionally override `%nbdev_default_export` by specifying a module: `%nbdev_export special.module`"""
```

without giving you a headache, I'd like to try to convince you that magic flags should be an exception (o:
a big part of the value of magic flags is the "in-notebook" help that they provide - if we limit to 1-line comments here, we could be limiting the value of the "in-notebook" help

I'm also hoping that other flags can have even longer comments - e.g. `nbdev_slow_test` would have
```
Put an `%nbdev_slow_test` magic on each "slow" test cell that you do not want to be run by default.
To apply this flag to all tests in a notebook, one cell should contain: `%nbdev_slow_test all`
These tests can be executed with the command line tool: `nbdev_test_nbs --flags slow`
```

what do you think?

sgugger:
Agreed for the magic command docs since that's the whole point of doing this :)
I've reviewed and apart from some minor formatting things (I like doc comments to begin with a capital for instance ;-) ) this all looks great to me. Thanks for adding more tests and being very thorough in!

I've tested this on our existing projects to test it doesn't break anything and it's actually better :) nbdev used to remove commented lines that are 0-indented (I'm just discovering this) but this PR fixes that!

sgugger:
PS: Do you have a Twitter account? I'm going to show the magic commands there and would like to credit your hard work.

pete88b:
Thanks Sylvain but I've never used twitter (o:

borisdayma:
That's an awesome functionality you're building! Congrats!

Issue #138, opened 2020-04-24
nbdev_new fails on Windows


epogrebnyak:
Should  `nbdev_new` be able to create a project template in a new folder?

I'm running this command in on Windows and get `Access denied` error

```
> nbdev_new pkg2

Traceback (most recent call last):
  File "d:\anaconda3\lib\runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "d:\anaconda3\lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "D:\Anaconda3\Scripts\nbdev_new.exe\__main__.py", line 9, in <module>
  File "d:\anaconda3\lib\site-packages\fastscript\core.py", line 73, in _f
    func(**args.__dict__)
  File "d:\anaconda3\lib\site-packages\nbdev\cli.py", line 258, in nbdev_new
    if os.path.isdir(path): shutil.rmtree(path)
  File "d:\anaconda3\lib\shutil.py", line 513, in rmtree
    return _rmtree_unsafe(path, onerror)
  File "d:\anaconda3\lib\shutil.py", line 392, in _rmtree_unsafe
    _rmtree_unsafe(fullname, onerror)
  File "d:\anaconda3\lib\shutil.py", line 392, in _rmtree_unsafe
    _rmtree_unsafe(fullname, onerror)
  File "d:\anaconda3\lib\shutil.py", line 392, in _rmtree_unsafe
    _rmtree_unsafe(fullname, onerror)
  File "d:\anaconda3\lib\shutil.py", line 397, in _rmtree_unsafe
    onerror(os.unlink, fullname, sys.exc_info())
  File "d:\anaconda3\lib\shutil.py", line 395, in _rmtree_unsafe
    os.unlink(fullname)
PermissionError: [WinError 5] Отказано в доступе: 'C:\\Users\\Евгений\\Desktop\\notebooks\\pkg2\\.git\\objects\\pack\\pack-32fff01fcfc3d5824e4097a65e36e8abd1bcd22c.idx'
```

sgugger:
I have not tried to use nbdev on Windows so I don't know if this command works. It seems like your python process does not have permission to remove a folder, which you could do manually.

epogrebnyak:
Maybe shutil.rmtree cannot remove a hidden file on Windows. Perhaps safer to use
`ignore_errors=True` below?

https://github.com/fastai/nbdev/blob/master/nbdev/cli.py#L258

[...]

PR #155, opened 2020-05-18
nbdev_new fails on Windows #138

pete88b:
Fix for https://github.com/fastai/nbdev/issues/138
On windows, .git is created with a read-only flag; so `rmtree(path/".git)"` fails - and then in the cleanup section `rmtree(path)` fails for the same reason.
I put the cleanup `rmtree(path)` in a try/except so that the real cause of the problem doesn't get lost in a stack trace if cleanup fails.


sgugger:
Thanks for your PR, it's looking good but can you please adjust to the current style of nbdev which is to have 1-liner for doc comments (and jsut double quotes, no triple double quotes) for a short summary of what the function does and the rest of the documentation in the notebook only? Here it's `rmtree_onerror` doc comment that should be changed.

pete88b:
done - I realized it wasn't a good comment (just writing what the code was doing) - hope this is better

sgugger:
That's how we prefer it, thanks :)

Issue #156, opened 2020-05-18
hide_output in README.md


lgvaz:
In `index.ipynb` I marked some cells with #hide_output, but the annotation is not correctly being parsed, the current results at `README.md` is a follows:

---
```python
#hide_output  
code here...
```
Output here

---

sgugger:
I don't think `hide_output` flags are supported in the conversion to Markdown. Will have look when I have time.

pete88b:
this might be really simple - because `convert_nb` already calls `hide_cells` and `collapse_cells` via `process_cell` ...
we could
- create hide-md.tpl (as a copy of hide.tpl) - change hide-md.tpl to extend from markdown.tpl
- update jekyll.tpl and md.tpl to extend from hide-md.tpl

and get all of the hide/collapse behaviour.

If this sounds OK, I can do some proper testing and create a PR

sgugger:
Sure, you can go ahead and test if you want to tackle this one.

pete88b:
I think I have a fix https://github.com/fastai/nbdev/pull/157

sgugger:
Should be fixed by #157

PR #157, opened 2020-05-20
fix issue #156

pete88b:
some details that might help review
- hide-md.tpl is a copy of hide.tpl - with a few differences
    - the zero indent of the `details` tags is important - they won't render properly if we indent like in hide.tpl
    - we need {{ '' }} after </details> or the next part (usually some output) is not rendered correctly
    - I removed the `{% block output_area_prompt %}` section - as this comes from basic.tpl
    - I removed the `{% block codecell %}` section - this caused "{% raw %}" to be rendered in the md

### DIFF
--- a/nbdev/templates/hide-md.tpl
+++ b/nbdev/templates/hide-md.tpl
@@ -1,0 +1,33 @@
+{%- extends 'markdown.tpl' -%}
+
+{% block input_group -%}
+{%- if cell.metadata.collapse_show -%}
+<details class="description" open>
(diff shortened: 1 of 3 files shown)


sgugger:
@hamelsmu This might impact you on fastpages, so I'd like your review before merging.

hamelsmu:
Thanks for tagging me @sgugger AFAIK this will not affect fastpages, because we are not using this conversion path to markdown at all.  I believe it is safe to merge.  

LGTM

PR #158, opened 2020-05-20
Create binder tags for PRs

hamelsmu:
[mybinder.org](https://mybinder.org/) is a free service maintained by the Jupyter community that allows you to serve Jupyter Notebooks with the right dependencies for free.  Furthermore, dependencies can be targeted on a branch or a pull request.   

This Action will drop this badge on all PRs which is a link to view a notebook on the branch of the PR, which will have dependencies preinstalled according to whatever is in `environment.yml`.  The Action will make a comment that looks like this:

image

@sgugger let me know if this is something you are interested in.  I have found it very helpful for my projects.  cc: @betatim

### DIFF
--- a/.github/workflows/binder.yaml
+++ b/.github/workflows/binder.yaml
@@ -1,0 +1,29 @@
+name: Binder
+on: 
+  pull_request:
+    types: [opened, reopened]
+
+jobs:
+  Create-Binder-Badge:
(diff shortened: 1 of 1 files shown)


hamelsmu:
eh ... let me close and do some more testing on Pull Requests

PR #159, opened 2020-05-24
adding %nbdev_export_and_show and %nbdev_export_internal

pete88b:
here's a couple more, whenever you're ready (o:

Hopefully the changes to clean_exports and remove_hidden make sense - here's what I was thinking;
- no flags, except hide, should make a difference to markdown
- magic and comment flags should do the same things
- flag behavior should be consistent. e.g. #exporti bad mod -> won't get picked up for export export, so it shouldn't be hidden


sgugger:
This all sounds good to me, thanks!

PR #162, opened 2020-05-27
adding magic test flags

pete88b:
I've made a small change to current behavior: "all" test flags now have to be configured in settings.ini - i.e. if you don't have the flag configured in `tst_flags`, it won't get picked up as an "all" test flag.

It feels a little strange that you can have any number of tests flags in a cell, but only one "all" flag per notebook.
Can I suggest we allow any number of "all" test flags? If yes, let me know if I should update this PR or create a separate one.


sgugger:
The change sounds good to me. We can definitely allow more than one "all" flag, I just never went down the road of coding it since I never needed it ;)
If you want to address that in a separate PR, go ahead!

PR #164, opened 2020-05-28
allow multiple all test flags

pete88b:
Hope you don't mind but I've also removed
`mod = find_default_export(nb['cells'])` from test_nb
mod isn't used and find_default_export has no side effects.


sgugger:
Ah yes, it's probably some code I forgot to remove when I was adding a cell that imports that module. Thanks a lot!

Issue #154, opened 2020-05-18
show_doc and `# exports module` don't play nice


BlackHC:
Using nbdev  0.2.18.

I added a module to an #exports to route code somewhere else. Now `nbdev_build_docs` fails with 

```
An error occurred while executing the following cell:
------------------
show_doc(BayesianCNN, default_cls_level=2)
------------------
```

(which is a cell that is created by the html conversion code for the cell whose target module I've changed)

I hope this provides sufficient information to reproduce it.

(BayesianCNN is a class that I defined in that cell as example, and then I thought: maybe I should collect this model in its own module to avoid copy-pasting the code over my notebooks.)

Thanks,
 Andreas

sgugger:
There is nothing we can do to help without:
1. seeing the whole stack trace
2. seeing the notebook that creates the problem.

We are not magicians :)

BlackHC:
```
nbdev_build_lib
Converted 01_batchbald.ipynb.
Converted 02_joint_entropy_computation.ipynb.
Converted 03_consistent_mc_dropout.ipynb.
Converted 04_active_learning.ipynb.
Converted 05_repeated_mnist.ipynb.
Converted 06_example_experiment.ipynb.
Converted 07_black_box_training.ipynb.
Converted A1_some_models.ipynb.
Converted index.ipynb.
touch batchbald
nbdev_build_docs
converting: /home/blackhc/PycharmProjects/bald-ical/03_consistent_mc_dropout.ipynb
An error occurred while executing the following cell:
------------------
show_doc(BayesianCNN, default_cls_level=2)
------------------

---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
<ipython-input-5-0dff4c94e4b1> in <module>
----> 1 show_doc(BayesianCNN, default_cls_level=2)

NameError: name 'BayesianCNN' is not defined
NameError: name 'BayesianCNN' is not defined

Traceback (most recent call last):
  File "/home/blackhc/anaconda3/envs/active_learning/bin/nbdev_build_docs", line 8, in <module>
    sys.exit(nbdev_build_docs())
  File "/home/blackhc/anaconda3/envs/active_learning/lib/python3.7/site-packages/fastscript/core.py", line 73, in _f
    func(**args.__dict__)
  File "/home/blackhc/anaconda3/envs/active_learning/lib/python3.7/site-packages/nbdev/cli.py", line 101, in nbdev_build_docs
[...]

PR #165, opened 2020-05-29
fix issue #154

pete88b:


sgugger:
Thanks, this looks very good to me!

PR #168, opened 2020-05-31
fixed issue 167

pete88b:
split_flags_and_code will behave as nbdev used to if no magic flags are found (i.e. remove just the 1st line) but we need different logic to remove multiple magic flags etc.
This means that comment flags and magics behave differently )o: I think this simplification is OK - but if you want comments/magics to be the same, please let me know


sgugger:
I don't expect to be multiple flags in an exported cell (what would be the second flag?) It generally happens in test cells.
So I guess, this is okay. Thanks for the fix!

PR #170, opened 2020-06-02
added %nbdev_collapse_input/output

pete88b:
as well as adding magic flags I have
- updated customstyles.css to avoid "Show Code" overlapping output. e.g. https://nbdev.fast.ai/export2html/#Collapsable-Code-Cells
- created an `nbdev_upgrade` command that will update customstyles.css for existing nbdev projects
    - if you create an nbdev project from https://github.com/fastai/nbdev_template/, `#collapse` flags do not give you show/hide behavior in the HTML docs

The following animation starts with customstyles.css from nbdev_template, then
- adds `details.description[open] summary::after` to make "Show Code" visible
- adjusts `margin-top` to avoid overlap (Note: we still get some overlap with "Hide Output")
- styles the "Show Code" text to look clickable and moves it over to the right

collapse-style

You can see this test page at https://pete88b.github.io/decision_tree/test_doc/
Note: we have some overlapping output in the last section - I can deal with that in a separate PR


pete88b:
@hamelsmu hopefully this makes more sense than the message I sent (o:

sgugger:
This all looks good to me, and if @hamelsmu does a thumbs up for the custom css, I assume it does not break fastpages :)
Thanks a lot!

PR #173, opened 2020-06-04
fix issue 171

pete88b:
Hope you're ok with this simple approach. I tried a few other things but they ended up getting pretty complicated when trying to handle imports that span multiple lines etc

Note: I needed to make `_re_show_doc` a little more specific before using it in `ExecuteShowDocPreprocessor` - otherwise it picks up too many things - this is probably better for `hide_cells` too


pete88b:
can we/you update the docs in github too?

sgugger:
LGTM, thanks for working on this!

PR #174, opened 2020-06-08
introducing %nbdev_hide

pete88b:
In `07_clean.ipynb` we have: `cell_metadata_keep = ["hide_input"]` Can I check that we don't need to add "hide_output", "collapse_hide" etc because "hide_input" is the only one that can be set via jupyter extension?

Couple of notes that might help:
- `_mk_flag_re` was changed to match "any number of spaces and/or tabs" after flag - rather than "any number of whitespace" - so that we don't remove line-breaks after flags when migrating.
- fixed a mistake in `check_re_multi` - it was ignoring the `code_only` param


nbs/tutorial.ipynb
```diff
@@ -191,14 +191,24 @@
     "\n",
```
sgugger, line 191: <p>I don't understand this part: we want the cell hidden in the docs, that's why we use #hide.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/fastai/nbdev/pull/174/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='21'/>

sgugger:
This looks good to me, I just have a question in one of your changes on 00_core.ipynb.

In 07_clean, we should keep all relevant metadata. Since "hide_output" and "collapse_hide" are only inferred when we build the docs, we don't need to keep them (they are never saved with the notebook in any case).

nbs/tutorial.ipynb
pete88b: I think ReviewNB might be making this hard to read. In jupyter it should look like:

image

and in edit mode:

image

part of the change I just made means that both markdown and code cells will be hidden, if they contain a `#hide` flag.
In this case, we want the code example with `#hide` to show up in the docs, so we use the HTML entity `&#35;` - because `&#35;hide` won't get picked up by `_re_hide`.
If we're using HTML entities, we need to use pre tags rather that triple back-ticks.

make sense?

sgugger: Got it!

sgugger:
Thanks a lot!

PR #180, opened 2020-06-12
added nbdev_default_class_level

pete88b:
I've fixed a couple of problems in this PR as well as adding the new magic.

- **03_export2html.ipynb** `ExecuteShowDocPreprocessor` will no longer run cells containing a `notebook2script()` call
    - can't believe it took me this long to work out why build docs failed so often (o:
- **00_export.ipynb** `export_names` will no longer add python keywords to `__all__`
    - I put https://github.com/pete88b/decision_tree/blob/master/72_if__name__.ipynb together to answer a forum question and then got errors because 'except' was in `__all__` https://github.com/pete88b/decision_tree/blob/master/decision_tree/test/expif.py 

Hope putting these things together is easier/quicker for you that having 3 separate PRs

Any thoughts on magic flags for `_all_` and `show_doc` before I start work on them?


sgugger:
LGTM and good job for those two additional fixes :)
Thanks a lot!

PR #182, opened 2020-06-19
Nbdev add2  all  

pete88b:
Hope you like these ideas. Here's a couple of notebooks to show how [nbdev_show_doc](https://github.com/pete88b/decision_tree/blob/master/51_test_show_doc.ipynb) and [nbdev_add2__all__](https://github.com/pete88b/decision_tree/blob/master/61_test_add2__all__.ipynb) can be used.

Note: We don't migrate `show_doc` to `%nbdev_show_doc` as part of the update - please let me know if you have a preference for this to be included.

I've also fixed a little bug in 02_showdoc.ipynb that meant enums were being displayed as classes
```
-        if is_enum(elt.__class__): name,args = _format_enum_doc(elt, qname)
+        if is_enum(elt):           name,args = _format_enum_doc(elt, qname)
```

couple of questions
- would you like me to "upgrade" nbdev to use magic flags?
    - i'd love to see nbdev using magics (o:
- fastai2 seems to duplicate nbdev source - would now be a good time to clean this up?
    - i'm not sure if anything in `fastai2/notebook` is used
    - i'd be happy to help "upgrade" fastai2 to use magic flags too


nbdev/export.py
```diff
@@ -9,6 +9,7 @@
 from .imports import *
 from fastscript import *
 from keyword import iskeyword
+from .flags import parse_line
```
sgugger, line 12: Importing a notebook that is numbered after is dangerous, could you move that function either in this notebook or in imports?

pete88b: I think this is safe because flags.py is written the old fashioned way (i.e. not exported from 08_flags.ipynb).
Let me know if I'm missing something and I'll move parse_line to imports.

pete88b:
Can I check that you're happy with the name `nbdev_add2__all__ `? ... it seems `nbdev_add2all` might be more natural for some.
I chose `nbdev_add2__all__ ` to be explicit but I find dunders a bit ugly - and I can't help thinking that a shorter name might be better.

nbdev/export.py
sgugger: Oh, had missed that. I think the notebook should be renamed flag_tests, to make this clearer.

sgugger:
Now that you mention it, I prefer `nbdev_add2all`. It's clear enough if someone is looking for it with tab complete and the dunder do look weird.

nbdev/export.py
pete88b: done

pete88b:
think this one is good to go now

sgugger:
Thanks a lot!

PR #187, opened 2020-07-10
added tags to cell_metadata_keep

abbscadaminds:
This is to allow the usages of tags in notebooks to also be in the repo when using the nbdev githooks.


sgugger:
I don't see why not, thanks!

PR #192, opened 2020-07-28
Update customstyles.css

### LINKED ISSUE #160: Deal with horizontally overflowing cell output in GitHub Pages
I found it necessary to handle overflow behaviour [not just for links](https://github.com/fastai/nbdev/blob/e9a5b1f76677d09da239be640f1cb8d179994a31/docs/css/customstyles.css#L88-L104) but for output_html in general. Perhaps the default for `customstyles.css` could be like so:

~~~
.output_html {
    overflow-y: scroll;
}
~~~

Otherwise, the static github.io page can trivially break, also wrt. the sidebar menu.

hamelsmu:
Fixes #160

### DIFF
--- a/docs/css/customstyles.css
+++ b/docs/css/customstyles.css
@@ -1329,3 +1329,7 @@ div.input + details {
     /* show/hide output */
     margin-top: -25px;
 }
+
+.output_wrapper{
+    overflow: scroll;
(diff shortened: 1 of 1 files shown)


jph00:
Merci :)

PR #193, opened 2020-07-28
Fix url construction

hamelsmu:
@jph00 I thought maybe you fixed this yesterday, but this corrects the issue of URLs resolving to the proper location


jph00:
Guess I forgot to push!

PR #194, opened 2020-07-28
fixed add2all sync issue

pete88b:
Sorry for not spotting this before the latest push to pypi but I found a couple of issues today.

1) we were only looking for `%nbdev_add2all` in the code section of the source - so the following was not getting picked up

```python
%nbdev_export
%nbdev_add2all ["_not_included_by_default"]
```

2) using `_all_` or `%nbdev_add2all` in a cell with other code worked fine when going nb->lib - but the `_all_` was getting lost when syncing lib->nb.

The following code in a nb:

```python
%nbdev_export
def _not_included_by_default(): pass
_all_ = ["_not_included_by_default"]
```

was getting written to lib as:

```python
%nbdev_export
def _not_included_by_default(): pass
```


jph00:
Thanks for noticing.

jph00:
@pete88b this seems to break nbdev_update_lib in a subtle way - it causes the string `nbdev_comment` to be removed even when it's actually needed. Specifically, I now can't use that command on nbdev itself, since that string appears in the code!

pete88b:
Hi @jph00, you're right of course but this is expected: When I made this change, I couldn't find "nbdev_comment" in any fastai code - so I thought it'd be safe to reserve the comment "#nbdev_comment ".
Can you share your code that needs "#nbdev_comment "?
Maybe the answer will be to use a more complex comment like "#nbdev_comment!DO_NOT_REMOVE "

jph00:
That's OK I've worked around it now. The issue is that the code you wrote in nbdev contains `nbdev_comment` itself :).

pete88b:
Hi @jph00, you're right of course but this is expected: When I made this change, I couldn't find "nbdev_comment" in any fastai code - so I thought it'd be safe to reserve the comment "#nbdev_comment ".
Can you share your code that needs "#nbdev_comment "?
Maybe the answer will be to use a more complex comment like "#nbdev_comment!DO_NOT_REMOVE "

pete88b:
sorry @jph00 - I've got you now:
`nb['cells'][i]['source'] = flags + '\n' + c.replace('#nbdev_comment ', '')` turns itself into
`nb['cells'][i]['source'] = flags + '\n' + c.replace('', '')`

this should not `replace` but remove from the start of the line only.

really sorry to have wasted your time with this - sounds like you've got it covered but please let me know if you need me to do anything

jph00:
All good thanks :)

Issue #185, opened 2020-07-08
Handling of __future__ imports


davidallen353:
I typically write python in Python3.7, but I like to use the new annotations described in PEP-563 (planned for 3.10).

To do this, I include the future import `from __future__ import annotations`. However this must be the first executable line of code.

As far as I can tell, there is not a way of telling nbdev to include an exported line before the `__all__ = []` statement. Thus, the compiled code cannot run. Specifically, it raises the follow exception on import `SyntaxError: from __future__ imports must occur at the beginning of the file`.

I am not sure how best to fix this. I did write in a work-around. Specifically, I modified the `create_mod_file()` function in `nbdev.export` to read

```python
def create_mod_file(fname, nb_path):
    "Create a module file for `fname`."
    fname.parent.mkdir(parents=True, exist_ok=True)
    file_path = os.path.relpath(nb_path, Config().config_file.parent).replace('\\', '/')
    with open(fname, 'w') as f:
        f.write(f"# AUTOGENERATED! DO NOT EDIT! File to edit: {file_path} (unless otherwise specified).")
        f.write('\nfrom __future__ import annotations')  # THIS LINE IS THE WORK AROUND
        f.write('\n\n__all__ = []')
```
This is admittedly a very hack-y workaround, but it meets my needs.

pete88b:
I can think of a couple of ways that nbdev could accommodate this;
- change export to write cells that contain `from __future__ ...` before `__all___`
    - sync would also need to be updated as cells might be written out of order
    - +ve: it would "just work" for nbdev users / -ve: quite a lot of changes to nbdev
- add a callback to [nbdev_callbacks](https://github.com/fastai/nbdev/pull/188) and update `create_mod_file` so users can provide addition header text
    - +ve: really simple to implement (assuming PR 188 is accepted) / -ve: users have to put `__future__` imports in 2 places (notebooks and callback)

Anyone else have any ideas? happy to work on these changes either way (o:

jph00:
I think we should try the first way, since it really ought to "just work". Maybe try a PR and see how hairy it is? If it ends up making the code significantly more complex, we might want to revert and use the callback idea instead.

PR #195, opened 2020-07-29
support __future__ imports

pete88b:
This PR follows on from PR https://github.com/fastai/nbdev/pull/194 (i.e. uses the `#nbdev_comment` prefix) and adds support for `__future__` imports (as mentioned in Issue https://github.com/fastai/nbdev/issues/185).

If it helps, here's and [example nb](https://github.com/pete88b/decision_tree/blob/master/74_from_future_import.ipynb) and [module](https://github.com/pete88b/decision_tree/blob/master/decision_tree/test/from_future.py) that use `__future__` imports.


pete88b:
Just found that when using `__future__` imports, `nbdev_test_nbs` fails with [python 3.6](https://github.com/pete88b/decision_tree/runs/928572052?check_suite_focus=true) but works fine with [python 3.7](https://github.com/pete88b/decision_tree/runs/928758707?check_suite_focus=true) - would you expect this kind of detail to be included in the docs?

jph00:
Thanks for this. Yes generally we expect to target py36+. So things failing on py36 should either be fixed, or if they're minor optional bits of functionality, at least noted.

Issue #209, opened 2020-08-10
Run on Colab Badge


hamelsmu:
See [this discussion](https://discordapp.com/channels/689892369998676007/689892370002870284/742526196377583696).

I will take on this issue.

PR #210, opened 2020-08-11
Add ability to have Colab badges on pages

hamelsmu:
EDIT: this is ready for review now

This closes issue #209


nbdev/templates/jekyll.tpl
```diff
@@ -7,6 +7,7 @@ sidebar: home_sidebar
 {% if resources.tags != "" and resources.tags != nil %}tags: {{resources.tags}}{% endif %}
 {% if resources.summary != "" and resources.summary != nil %}summary: "{{resources.summary}}"{% endif %}
 {% if resources.summary != "" and resources.summary != nil %}description: "{{resources.summary}}"{% endif %}
```
hamelsmu, line 9: @jph00 @sgugger  I noticed this line of code while I was doing this PR.  Looks like a bug, WDYT?  should the value should be description, not the summary again, right?

jph00:
Nice! I think we need a way to turn them on and off per page. If on for the site, it shouldn't be added to pages that aren't from notebooks.

And needs some docs including in readme

hamelsmu:
@jph00 I made the following changes:

1. Edited `index.ipynb` to add a section about badges
2. Built the docs which rendered `README.md` as well as all of the HTML pages with the new Colab Badge
3. Added the ability to turn off a badge selectively on a page even though the site-wide option is turned off via the `hide_colab_badge` front matter, this is what is used in fastpages, BTW.
4.  Re: ignoring non-notebook posts I will comment on the code that facilitates this naturally. 

LMK what you think

docs/_layouts/page.html
```diff
@@ -5,6 +5,9 @@
 <div class="post-header">
     <a id="{{page.title}}"></a>
     <h1 class="post-title-main">{{ page.title }}</h1>
+    {% if page.nb_path %}
```
hamelsmu, line 8: @jph00 this line of code prevents non-notebook documents from getting a Colab badge

Issue #200, opened 2020-08-03
the /flags link is broken


hamelsmu:
The link [https://nbdev.fast.ai/flags](https://nbdev.fast.ai/flags) is broken on this page

![image](https://user-images.githubusercontent.com/1483922/89213051-5a3e1a80-d579-11ea-80d0-1180654e34d1.png)

hamelsmu:
I would try to fix this but I still havent' been able to resolve my dependency issues with nbdev :(

pete88b:
This could be a little tricky ...
nbdev is converting flags (surrounded by back-ticks) to a link to the doc page for the flags module but ... in this case
- we just want "flags" to be formatted as code: because it's a parameter name

We've used flags as a parameter name in a few notebooks so I think we'll find a few more broken links like this.

I'm away for a couple of weeks but I'm happy to take a look if no-one else picks this up before I get back

jph00:
I'll happily leave it for you @pete88b :) Let me know if you have any questions.

PR #218, opened 2020-08-20
updates to how backtick names are converted to doc links

pete88b:
Hi @jph00 

this one takes care of https://github.com/fastai/nbdev/issues/200 by
- checking that a module name is also a valid doc name before linking
- not converting parameter names to links

few other things
- should "isnter" be "enter" in the following?
    - "Search for doc links for any item between backticks in `text` and isnter them"
- should we write a map of module to notebook names to _nbdev.py - so `doc_link` doesn't have to assume these names correspond 
- the "not converting parameter names to links" idea works for show_doc calls for now - would you like to extend this behavior to markdown following a show_doc call too?


jph00:
Great. Yes `isnter` is wrong. What problem would "write a map of module to notebook names" solve? What pros and cons do you see of "extending this behavior to markdown following a show_doc call"?

pete88b:
@jph00 please see issues https://github.com/fastai/nbdev/issues/220 and https://github.com/fastai/nbdev/issues/221

PR #226, opened 2020-08-31
Stricter pattern matching for imports

feynlee:
`_re_import` now has stricter matching conditions for imports, so that it will not match `from xxxx` without the following `import`. This prevents code blocks that include things like
 
```py
query="""
select *
from database
"""
```
 
from running when generating docs due to the misclassification of it being imports. 

This can be a conner case, but the stricter condition would not miss any legitimate imports so it shouldn't hurt anything.


jph00:
Thanks!

PR #229, opened 2020-09-02
Adding google colab cell tags to cell_metadata_keep

matjesg:
"colab_type" and "cellView" in cell metadata allow google colab to define cell type and view, e.g.,  "cellView": "form",
and "colab_type": "code" allow to completely hide a code cell and only show markdown in google colab


matjesg:
Example:
image

jph00:
Unfortunately that's going to cause big diffs and conflicts in git if we keep those in, since some folks use colab and some don't

PR #233, opened 2020-09-06
Update README.md

kevinbird15:
Fixes a small issue where README doesn't render properly


jph00:
This actually needs to be fixed in index.ipynb - readme is autogenerated. I'll fix it there now.

PR #234, opened 2020-09-06
adding show_source option to show_doc

kevinbird15:
This is an option that seemed like it would have been useful when writing a blog post and nbdev's show_doc seemed like a good spot to put it.  This gives the user the option to show_source when they do a show_doc.  By default this will be disabled so it shouldn't break anything else unless somebody was using `_format_func_doc` or `_format_cls_doc` directly.


jph00:
It's a cool idea, but not something I want to add to the library. (Amongst other things, it relies on get_ipython being present, and also relies on internal details of ipython that may change.)

kevinbird15:
Is this something you would want to add if it didn't use get_ipython or would it be better to just do it outside of nbdev?

PR #239, opened 2020-09-07
Fix: correct notebook2html path operation for Windows.

rickatx:
With the default project structure, the glob() in notebook2html() can easilly find a copy of a .ipynb file in .ipynb_checkpoints. In this case, running nbdev_build_docs will fail, assuming the .ipynb imports the project (the relative path is not correct from this dir.) The current code to skip hidden path components searches for '/.', whereas str(f) will return the Windows backslash path, so the search will fail. f.as_posix() returns a '/' delimited path on all platforms.

I verified failure of nbdev_build_docs before the fix, and success after installing the fix.


jph00:
Many thanks.

PR #242, opened 2020-09-09
Magic flags ipython extension

pete88b:
Hi @jph00 - do you think it would be nicer to load magic flags as an ipython extension? so you don't have to `import nbdev` to get magic flags.

I've removed `from nbdev import *` from `00_export.ipynb` to make use of the 3rd option described in `magic_flags.ipynb` - let me know if you like this approach and I'll update the other notebooks.
If you'd prefer not to use `ipython_config.py`, we could change the imports to `%load_ext nbdev.flags`.

While this works in colab ...
```
!pip install git+https://github.com/pete88b/nbdev.git@magic-flags-ipython-extension
%load_ext nbdev.flags
```
I don't think https://github.com/pete88b/nbdev_colab_helper is ready to be referenced in the nbdev docs yet

### DIFF
--- a/nbdev/flags.py
+++ b/nbdev/flags.py
@@ -116,4 +116,7 @@ if IN_IPYTHON:
     except: pass # do not fail if we can't find config
 
 def load_ipython_extension(ipython):
-    "This function makes `nbdev.flags` an ipython extension. No logic needed - we just need imports to run"
+    "You can `%load_ext nbdev.flags` to make magic flags, show_doc and notebook2script available in a notebook"
+    from .showdoc import show_doc
+    from .export import notebook2script
+    ipython.push(dict(show_doc=show_doc,notebook2script=notebook2script))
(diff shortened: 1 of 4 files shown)


jph00:
Is there any major benefit to this approach?

My general preference is to avoid making any changes, because this ipython magic functionality has taken up a huge amount of maintenance time for me, and I'm keen for that not to continue. (I'm really grateful for the hard work you've put in to this - it's just that I don't have a lot of time and I have to be careful of how I spend it, especially with everything that's going on at the moment).

FYI, I'm planning to replace magic flags with comments in fast.ai notebooks since that's what's been more convenient for me.

pete88b:
Hi @jph00  No major benefit but this tries to solve the problem that you can't even run a cell that contains them unless you've first imported a module.

Totally cool if you don't want these changes - I'm here to help and make things better if I can - but definitely don't want to be a drain on your time

jph00:
Thanks Pete - really appreciate it :)

PR #243, opened 2020-09-09
Enable Codespaces for nbdev

hamelsmu:
Similar to https://github.com/fastai/fastai/pull/2779


hamelsmu:
I'll go ahead and merge this as it isn't related to code directly is is more related to "GitHub" automation

PR #250, opened 2020-09-13
Updated nbdev to use 6.0 with pre-existing 5.x templates

MSeal:
I believe this fixes all the issues for nbconvert 6.0 usage.

I left the templates in 5.x style and used the template_file backwards compatibility supported path for referencing them to keep the PR minimal. NBconvert 6.0 introduced templates as a directories so that it can include config files and multiple jinja files that can be composed as a unit. https://nbconvert.readthedocs.io/en/latest/customizing.html has guides on how this works and how to extend it now if you wish to change to the new pattern.

`make test` complains about `06_cli.ipynb` where the id fields of the two values don't match. But this is happening for me on master with nbconvert 5.6.1 so I think it's unrelated.

`jupyter_client` was given a min version because it's caused issues for other libraries where the min version pin in nbclient isn't respected by pip in all scenarios due to it being a nested transitive dependency.

### DIFF
--- a/nbdev/export2html.py
+++ b/nbdev/export2html.py
@@ -510,14 +510,19 @@ def write_tmpls():
 
 # Cell
 def nbdev_exporter(cls=HTMLExporter, template_file=None):
-    cfg = traitlets.config.Config()
-    exporter = cls(cfg)
-    exporter.exclude_input_prompt=True
-    exporter.exclude_output_prompt=True
-    exporter.anchor_link_text = ' '
-    exporter.template_file = 'jekyll.tpl' if template_file is None else template_file
-    exporter.template_path.append(str(Path(__file__).parent/'templates'))
-    return exporter
+    # For backwards comaptability of 5.x style templates, we pass a full template_file path
+    # instead of manipulating template_paths
+    if not template_file:
+        template_file = 'jekyll.tpl'
+    if template_file and not os.path.isabs(template_file):
+        template_file = str(Path(__file__).parent/'templates'/template_file)
+    cfg = traitlets.config.Config(
+        exclude_input_prompt=True,
+        exclude_output_prompt=True,
+        anchor_link_text=' ',
+        template_file=template_file,
+    )
+    return cls(cfg)
 
 # Cell
 process_cells = [remove_fake_headers, remove_hidden, remove_empty]
--- a/nbdev/templates/hide-md.tpl
+++ b/nbdev/templates/hide-md.tpl
@@ -1,4 +1,4 @@
-{%- extends 'markdown.tpl' -%}
+{%- extends 'markdown/index.md.j2' -%}
 
 {% block input_group -%}
 {%- if cell.metadata.collapse_show -%}
--- a/nbdev/templates/hide.tpl
+++ b/nbdev/templates/hide.tpl
@@ -1,4 +1,4 @@
-{%- extends 'basic.tpl' -%}
+{%- extends 'classic/index.html.j2' -%}
 
 {% block codecell %}
     {{ "{% raw %}" }}
--- a/nbs/03_export2html.ipynb
+++ b/nbs/03_export2html.ipynb
@@ -1774,14 +1774,19 @@
    "source": [
     "#export\n",
     "def nbdev_exporter(cls=HTMLExporter, template_file=None):\n",
-    "    cfg = traitlets.config.Config()\n",
-    "    exporter = cls(cfg)\n",
-    "    exporter.exclude_input_prompt=True\n",
-    "    exporter.exclude_output_prompt=True\n",
-    "    exporter.anchor_link_text = ' '\n",
-    "    exporter.template_file = 'jekyll.tpl' if template_file is None else template_file\n",
-    "    exporter.template_path.append(str(Path(__file__).parent/'templates'))\n",
-    "    return exporter"
+    "    # For backwards comaptability of 5.x style templates, we pass a full template_file path\n",
+    "    # instead of manipulating template_paths\n",
+    "    if not template_file:\n",
+    "        template_file = 'jekyll.tpl'\n",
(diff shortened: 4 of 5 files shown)


jph00:
Gosh that's extraordinary generous of you to go to the trouble of fixing our code for us!

Thank you so much. I'm extremely grateful.

jph00:
Oh just realized we can't merge this until nbdev 6.0.2 is available on conda. Will revert and leave open.

jph00:
Huh - I guess there isn't a way to reopen PRs, unlike Issues. Nevermind - I know where to find the PR once conda packages are ready :)

MSeal:
Conda forge was updated yesterday as well: https://anaconda.org/conda-forge/nbconvert

jph00:
@MSeal I guess you're busy with parenting duty at the moment, so please don't feel any hurry or commitment to respond, but I thought I should let you know that I tried this patch and it didn't work out.  Running `03_export2html.ipynb` directly in Jupyter notebook causes a kernel error (or sometimes just uses 100% of CPU on all cores). When it gives an error, it's this (in the jupyter console):

```
Traceback (most recent call last):
  File "/home/jhoward/anaconda3/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/home/jhoward/anaconda3/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/jhoward/anaconda3/lib/python3.7/site-packages/ipykernel_launcher.py", line 16, in <module>
    app.launch_new_instance()
  File "/home/jhoward/anaconda3/lib/python3.7/site-packages/traitlets/config/application.py", line 664, in launch_instance
    app.start()
  File "/home/jhoward/anaconda3/lib/python3.7/site-packages/ipykernel/kernelapp.py", line 612, in start
    self.io_loop.start()
  File "/home/jhoward/anaconda3/lib/python3.7/site-packages/tornado/platform/asyncio.py", line 149, in start
    self.asyncio_loop.run_forever()
  File "/home/jhoward/anaconda3/lib/python3.7/asyncio/base_events.py", line 541, in run_forever
    self._run_once()
  File "/home/jhoward/anaconda3/lib/python3.7/asyncio/base_events.py", line 1771, in _run_once
    handle = self._ready.popleft()
IndexError: pop from an empty deque
```

Also, the exported docs don't contain any of the jekyll headers, or any HTML header.

jph00:
Running the same code as a script works, however - it's only when run inside a notebook that it fails.

MSeal:
Still waiting on baby arrival, so I'll try to help while that's pending :D

I think this is unrelated to nbconvert and is instead an issue with multiprocessing inside ipython / ipykernel with an open python bug:

https://github.com/jupyter/notebook/issues/5261 -> https://github.com/ipython/ipython/issues/12396 -> https://bugs.python.org/issue41041
and maybe this is the same thing (couldn't find an issue to tie to it): https://stackoverflow.com/questions/62041414/python-simple-loop-parallelization-jupyter-notebook

I tested with a pool of size 1 and it still hangs, but running the conversions sequentially (avoiding the `parallel` call) works fine. With a hacked together version of `parallel` using `from multiprocess import Pool` from the multiprocess package I was able to run the notebook inside Jupyter in parallel, which makes it look like it's indeed the linked issue and using the CPython package or a threadpool would get around the issue here.

jph00:
Oh that's cool that nbconvert doesn't hold the GIL so works with threadpool. I've made that change locally and it's working nicely! Thanks @MSeal 👍 

Any ideas why the exported docs don't contain any of the jekyll headers, or any HTML header? Perhaps it's a different base template I need to inherit from?

PS: I shared my top parenting tip in the first 5 mins of this interview - have a look because it might just change your life!... https://www.youtube.com/watch?v=t2V2kf2gNnI

MSeal:
Ahh so it does looks like I pointed to the wrong base in the PR for one of your templates -- change `{%- extends 'classic/index.html.j2' -%}` in `hide.tpl`  to `{%- extends 'classic/base.html.j2' -%}` to get the bare bones base template you were using before. See if that looks correct after that changes.

Thanks for the link!

jph00:
Thanks! :) Getting closer. Still missing the jekyll front-matter at the top. Any thoughts on how to make that re-appear? Or any general suggestions on where I should start looking?

jph00:
The value of `cfg` in `nbdev_exporter()` is:

    {'exclude_input_prompt': True, 'exclude_output_prompt': True, 'anchor_link_text': ' ', 'template_file': '/home/jhoward/git/nbdev/nbdev/templates/jekyll.tpl'}

So it's trying to use the jekyll tpl. That tpl contains:

```
{%- extends 'hide.tpl' -%}{% block body %}---
{% if resources.toc != "" and resources.toc != nil %}toc: {{resources.toc}}{% endif %}
{% if resources.title != "" and resources.title != nil %}title: {{resources.title}}{% endif %}
{% if resources.image != "" and resources.image != nil %}image: {{resources.image}}{% endif %}
{% if resources.hide_colab_badge != "" and resources.hide_colab_badge != nil %}hide_colab_badge: {{resources.hide_colab_badge}}{% endif %}
keywords: {{resources.keywords}}
sidebar: home_sidebar
{% if resources.tags != "" and resources.tags != nil %}tags: {{resources.tags}}{% endif %}
{% if resources.summary != "" and resources.summary != nil %}summary: "{{resources.summary}}"{% endif %}
{% if resources.summary != "" and resources.summary != nil %}description: "{{resources.summary}}"{% endif %}
{% if resources.nb_path != "" and resources.nb_path != nil %}nb_path: "{{resources.nb_path}}"{% endif %}
---
{% include 'autogen.tpl' %}

<div class="container" id="notebook-container">
        {{ super()  }}
</div>
{%- endblock body %}
```

Does that need to change at all?

PR #253, opened 2020-09-16
Removed callbacks

pete88b:
Hi @jph00 you OK with one PR to remove magic flags?
I can take the magic flags code out while disturbing remaining code as little as possible - i.e. I'm assuming this is just the 1st pass of a bigger refactor so I won't do any additional code cleanup yet.


jph00:
Thanks so much!

> Hi @jph00 you OK with one PR to remove magic flags?

Yup for sure :)

Issue #263, opened 2020-09-23
Feature Request: Allow for the doc heading and showing the source code


muellerzr:
Currently we have `exports`, which will show the function above the documentation heading, which looks a bit strange in some places. I'm not 100% how we could do this (though I will try to look into it and learn as I go), but my ideal scenario is we have the heading for the nbdev docstring followed by the source code of everything *after* the doc string. Something like below:

![image](https://user-images.githubusercontent.com/7831895/93957770-f9ce8c80-fd22-11ea-8361-819cc55cd1e7.png)

muellerzr:
Closing, need to rethink design more. May reopen if I decide on something

hamelsmu:
This is interesting.  I stared at this issue for a bit and questioned "why have source code in documentation?"   Then I kind of came around to the idea of having all source code in the documentation under a collapsible element by default.   That way, users can see source code in-situ really quickly without having to go to another web page. 

@jph00 WDYT?

@muellerzr sorry for crashing the closed isssue!  But you made me think!

muellerzr:
@hamelsmu perfectly fine! I was trying to wrap my head around it myself and couldn't quite figure out what format would look good, and I think having it hidden is a great solution! (potentially even with an option to default to open/closed? similar to other variations of hide)

hamelsmu:
Yeah I'm thinking under a collapsable element that is hidden by default so it doesn't clutter the docs (edited my comment above to clarify my intent).

hamelsmu:
BTW I found this behavior in fastcore, I also would expect the source code to appear below the heading.  

![image](https://user-images.githubusercontent.com/1483922/93964799-63e03500-fd15-11ea-9ff5-2fd7e50ad519.png)

hamelsmu:
ok just to clarify, now that I understand what is happening,  is to **always** show the source code under a collapsible heading that is closed by default, and thus do away with the need to to have both `#export` and `#exports` .... don't know if this is a good idea or not, just throwing it out there.  

I personally would enjoy always having the ability to see the source code in the same context because I'm lazy and since it can be hidden under a collapsible element, don't see that much downside of always including it.

jph00:
Personally, I don't find clicking a "source" link that opens a 2nd tab, and then tiling 2 tabs side by side, significantly harder than opening a collapsible div. At least, for something like `L` or `Learner` I think I'd rather have the option to show side-by-side like that.

[...]

PR #265, opened 2020-09-23
Place source code below heading on #exports

hamelsmu:
As discussed in #263 this makes it such that source code appears below the related heading instead of above the heading.  This is an example of how it looks 

image

Will do a separate PR for collapsible headings, but want to make sure this is what is desired first.   ﻿


jph00:
Super! Very thoughtful of you to include an image @hamelsmu :)

Issue #220, opened 2020-08-23
doc_link doesn't understand default export


pete88b:
nbdev allows module names to be different to notebook names via %nbdev_default_export but ... `doc_link` forces them to follow a specific pattern (if you want backtick links to your modules).

Please see [this demo page](https://pete88b.github.io/nbdev_demo/text_utils.html) for a more detailed explanation.

I'd really like to write a map of module to notebook names to _nbdev.py
- so doc_link doesn't have to assume that module and notebook names correspond and
- make `is_lib_module` work even when your project is not in `sys.path`

jph00:
I think I've come up with a pretty good way to create a system-wide mapping, that will also allow namespace extension modules to work. Basically, I think we might be able to leverage setuptools entry points for this functionality.

In other news, nbconvert 6 is out, and has broken pretty much everything :( I've pinned the version number for nbdev for now to work around this problem. When I do tackle the upgrade, I'm reasonably likely to remove the magic and callback functionality, since although it's extremely cool, it's beyond my ability to understand and maintain - so you might want to create a fork in you're interested in continuing to maintain that functionality.

pete88b:
Not great news about nbconvert 6 @jph00 )o: please feel free to at-mention me if you'd like to offload any of this work.

Makes sense to kick out magic flags and callbacks - if it turns out people miss them, I think both could be implemented without changes to nbdev (i.e. by pre-processing notebooks before nbdev processes them).

jph00:
Thanks so much :)

jph00:
@pete88b it'll be a while b4 I can look at this, so if you do happen to have the time and inclination to remove either or both of these, I'd be *very* grateful! (But please don't feel obliged, mind you...)

jph00:
@pete88b I'm going to start working on removing the magic flags now, so let me know ASAP if you've already made a start! :)

jph00:
OK magic flags removal is done now. (In master, not released)

PR #267, opened 2020-09-24
Fix inconsistent usage of %nbdev_export in the tutorial notebook

friethjof:
Hi, just a very minor change in `tutorial.ipynb`. I replaced `#export` with `%nbdev_export`, since the latter is referenced in the accompanying instructions.


friethjof:
I just read [Jeremy's comment](https://github.com/fastai/nbdev/issues/220#issuecomment-691021855) in issue #220. Since the magic functionality might be removed in the update to nbconvert 6 this change to the tutorial does not seem necessary :)

PR #268, opened 2020-09-25
Add example for specification of requirements to tutorial.ipynb

friethjof:
Hi,

I am proposing to change the instructions in `tutorial.ipynb` to specify the `requirements` in `settings.ini`.
Specifying the `requirements` in `settings.ini` as described in [install_requires in setuptools](https://packaging.python.org/discussions/install-requires-vs-requirements/#install-requires) leads to a parse error in the build within GitHub Actions.

The link given in `tutorial.ipynb` implies the following format:
```python
requirements = ['fastcore>=1.0.5', 'torchvision<0.7', 'matplotlib']
```
which would lead to a parse error. 

A format that does _not_ cause a parse error is
```python
requirements = fastcore>=1.0.5 torchvision<0.7 matplotlib
```

### DIFF
--- a/nbs/tutorial.ipynb
+++ b/nbs/tutorial.ipynb
@@ -704,7 +704,12 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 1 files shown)


jph00:
Thanks!

PR #272, opened 2020-10-03
Fix filter condition to match if filename is not 'index.ipynb'

butchland:
Hi, @jph00 
I think there might be a mistake in the filter condition. 
If not (and I just don't understand it)  just disregard this  PR.

Best regards,
Butch


jph00:
Thanks for the PRs. I'm doing some refactoring at the moment so will leave these open for a while - but don't worry I will get to them! :)

PR #273, opened 2020-10-03
Filter notebooks stored in ".ipynb_checkpoints" dirs from html genera…

butchland:
Hi, @jph00 

I encountered this problem because jupyter notebook creates these hidden directories `.ipynb_checkpoints` inside the
same dirs as where the notebooks are located. 

Jupyter then saves versioned copies of the main notebooks inside these directories and these are then named `xxxx-checkpoint.ipynb` where `xxxx.ipynb` is original notebook. 

This then cascades a file open error since `_get_title` will look for the generated html file name (generated by `_nb2htmlfname`) does not exist.

This problem doesn't show up on Github CI since `.ipynb_checkpoints` are git ignored and are never exported to github,
but it does show up if you call nbdev_build_docs in a local environment where jupyter notebooks have been saved with
checkpoints.

My fix is to just filter out those notebooks inside `.ipynb_checkpoints` directories.

Best regards,
Butch

### DIFF
--- a/nbdev/export2html.py
+++ b/nbdev/export2html.py
@@ -704,7 +704,7 @@ def _get_title(fname):
 def _create_default_sidebar():
     "Create the default sidebar for the docs website"
     dic = {"Overview": "/"}
-    files = [f for f in Config().nbs_path.glob('**/*.ipynb') if not f.name.startswith('_')]
+    files = [f for f in Config().nbs_path.glob('**/*.ipynb') if not f.name.startswith('_') and '.ipynb_checkpoints' not in f.parts]
     fnames = [_nb2htmlfname(f) for f in sorted(files)]
     titles = [_get_title(f) for f in fnames if 'index' not in f.stem!='index']
     if len(titles) > len(set(titles)): print(f"Warning: Some of your Notebooks use the same title ({titles}).")
--- a/nbs/03_export2html.ipynb
+++ b/nbs/03_export2html.ipynb
@@ -2358,47 +2358,6 @@
    "display_name": "Python 3",
    "language": "python",
    "name": "python3"
-  },
-  "language_info": {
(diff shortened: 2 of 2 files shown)


tcapelle:
You need to clean up the notebooks to pass the CI (nbdev_clean_nbs).
I would also filter `'checkpoints' not in f.parts` instead of the `.ipynb_checkpoints' not in f.parts`

butchland:
Hi @tcapelle,
I just checked --  running`nbdev_clean_nbs` only cleans the notebooks themselves. They don't remove the notebooks in the `.ipynb_checkpoints` directory themselves.  

Screen Shot 2020-10-07 at 11 29 26 PM

The problem is not the notebooks themselves -- its the fact that running `nbdev_build_docs` looks for the notebooks in the directories using patterns '**/*.ipynb' which  will include the checkpointed versions of the notebooks in the directory .ipynb_checkpoints (i.e. they will be named <name-of-original-notebook>-checkpoint.ipynb).

When you run `nbdev_build_docs` (even if you already ran nbdev_clean_nbs) this triggers an error.

Screen Shot 2020-10-07 at 11 36 15 PM

Also, the filter condition `checkpoints in f.parts`  will not match the notebooks inside `.ipynb_checkpoints` directory -- this is because f.parts is the list of path segments (not the whole path string) for the file e.g. /nbs/.ipynb_checkpoints/03_core.ipynb will have the f.parts ['nbs','ipynb_checkpoints'] so checking if `checkpoints` is in f.parts will not match, but looking for '.ipynb_checkpoints' will match f.parts.

I've also tested the fix on the same library and it fixes the problem. 

I also reran the nbdev_clean_nbs on my pr so that all checks now pass.

jph00:
Thanks!

Issue #260, opened 2020-09-18
The gitlab docs generator makes wrong sidebar links


jochym:
Since recent updates to nbdev (1.0.18 at least, probably earlier) the generation of sidebar navigation on gitlab stopped working correctly. It generates module_name.html links but the actual files are located in module_name/index.html. If you fix the paths and switch to custom sidebar it gets fixed in generation. If you keep autogenerated json file and patch it before running jekyll build, it still does not work (the links are still the same). You can check this in my fork of the gitlab template: (https://gitlab.com/jochym/nbdev_template). I have too little knowledge on jekyll to fix this or even properly debug but will gladly help by testing etc.

tcapelle:
I think it is also broken on github...
I updated the gitlab repo, let's try to fix this.

jph00:
I believe you just need to disable permalink: pretty in jekyll.

PR #274, opened 2020-10-03
Generate link without html prefix

butchland:
Hi @jph00,

I encountered an error where the doc site generated the wrong links in the sidebar to the subpages in the main site.

The new subpages are now created in the subfolders with the file 'index.html'.  (see `docs/_site` to 
see what I mean).

My fix removes the html suffix to the link  and creates the correct link to the folder for the document 
sub pages.

Best regards,
Butch


butchland:
I think this is a fix to #260 as well.

tcapelle:
it probably is, gitlab for instance does not like links with the .html extension. I am using custom sidebar for this.

jph00:
I believe you just need to disable `permalink: pretty` in jekyll.

PR #282, opened 2020-10-13
Fix for CWE-1022 CWE-200

byteshiva:
Fix for CWE-1022 CWE-200

refer:
https://github.com/fastai/fastpages/pull/433


hamelsmu:
Just some more background, I don't really have the expertise to evaluate this PR, perhaps someone else does.

PR #284, opened 2020-10-18
Fix bug where "IN_NOTEBOOK" is not imported.

kessido:
There appear to be a bug in the code where "IN_NOTEBOOK" is not imported... This now fix the issue.


jph00:
Thanks!

PR #298, opened 2020-10-31
Minimal example

### LINKED ISSUE #290: I've lost the thread
This is feedback about the documentation -- my goal is to let you know what roadblocks I'm hitting as a new user.  I hope this is helpful.

I followed the tutorial and tested everything using a minimal repo based on the template.  So far so good.  Now I am ready to add more material to the repo and try out additional features.  So that's what I'm looking for next.

But the next item after the tutorial is this page:

https://nbdev.fast.ai/export.html

I started reading that, and immediately I have lost the thread of what we are doing.  I don't understand what this page is telling me or why.  Is this what I need to know next?  Or is this information about the implementation that I don't need to know?  I can't tell.

The next item is

https://nbdev.fast.ai/sync.html

This is better.  I understand what this page is about, but I don't need it yet.  So, on to the next:

https://nbdev.fast.ai/showdoc.html

I think I understand this page, but it's not explicit about what problem it is solving, and I think I don't need it yet.

Sorry if this is too much detail -- my point is that as a beginner exploring this tool and deciding whether to continue, I have hit a stopping point after the tutorial.

The next steps I need help with are

1. Now that I have a minimal repo working, how do I get from here to doing real work?

2. Rather than starting a new project from scratch, what I really want to do is fit an existing project into this framework, so I could really use documentation for that.

3. I am actually considering using nbdev to generate a book rather than a Python library, so I would like to find documentation for that.

This looks like a really useful project, and I understand that writing documentation is hard.  So I hope this feedback is more helpful than annoying :)

hamelsmu:
This will close #290


hamelsmu:
@jph00 this is ready for review.  Since this is a documentation page that doesn't generate any code, only docs I went ahead and deployed the code in this PR to netlify so you can see what it looks like.  You can see a live docs preview of [this branch here](https://5f9e1f8fa392bfead2df40f3--nbdev-docs-preview.netlify.app/example.html).

_P.S. I have Netlify previews as part of the CI for fastpages which we can incorporate here if you like sometime_

PR #302, opened 2020-11-02
Update minimal example to close #301

### LINKED ISSUE #301: Improve Minimal Example Tutorial
- one thing to mention might be that you don't need to add "show_doc" for normal functions (non-methods), since it is auto-generated where you define the function
- and maybe explain why we don't auto-generate for methods, but instead want the dev to add prose/examples etc

ref #290

hamelsmu:
closes #301


hamelsmu:
@jph00 addressed your feedback in this PR (merging).

PR #304, opened 2020-11-04
fix docs link bug

hamelsmu:
**I checked all the filenames in fastai/fastai and this fix would fix the following bugs.  All other URLs stay exactly the same.**

NEW URL | OLD URL
-- | --
https://docs.fast.ai/torch_core | https://docs.fast.ai/torch.core
https://docs.fast.ai/tutorial.medical_imaging | https://docs.fast.ai/tutorial.medical.imaging
https://docs.fast.ai/test_utils | https://docs.fast.ai/test.utils
https://docs.fast.ai/pytorch_doc | https://docs.fast.ai/pytorch.doc

_P.S. I'm not sure the best way to add an additional test for this, as I would have to either import fastai which I don't want to add that dependency to nbdev, and to create some kind of new fake code that I export which I don't want to do either._

### DIFF
--- a/nbdev/showdoc.py
+++ b/nbdev/showdoc.py
@@ -284,7 +284,7 @@ def get_doc_link(func):
         nbdev_mod = importlib.import_module(mod.__package__.split('.')[0] + '._nbdev')
         try_pack = source_nb(func, mod=nbdev_mod)
         if try_pack:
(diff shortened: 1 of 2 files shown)


hamelsmu:
cc: @muellerzr

This closes https://github.com/fastai/fastai/issues/2876 and https://github.com/fastai/fastai/issues/2872

jph00:
Nice use of `partition`! :)

PR #313, opened 2020-11-09
Added support for imports with custom spacing

kessido:
current nbdev parsing does not handle cases where the import are as follow:
```
from lib.a    import *
from lib.bla import c
```
The reason being that the custom spacing after `lib.a` is not supported in the regex.

Added multi-space to the regex, and added test for verification.

### DIFF
--- a/nbdev/sync.py
+++ b/nbdev/sync.py
@@ -79,17 +79,17 @@ def relimport2name(name, mod_name):
 
 # Cell
 #Catches any from .bla import something and catches .bla in group 1, the imported thing(s) in group 2.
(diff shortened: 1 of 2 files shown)


jph00:
Thanks :) FYI nbdev v2 uses ast.parse so won't have issues like this any more.

kessido:
> Thanks :) FYI nbdev v2 uses ast.parse so won't have issues like this any more.

where can I find this so called "nbdev v2" P:?

jph00:
It's under development in the `nbconvert6` branch.

PR #331, opened 2020-12-03
Make install uses pip to install package

zonca:


jph00:
We generally recommend people use `pip install -e .`

Since that creates an *editable install*, there's no need to run it more than once (unless you add more entrypoints). So I don't see that there's really a reason to have `make install`.

Feel free to reopen this if I'm missing something!

Issue #357, opened 2020-12-30
nbdev searches for .ipynb in all subdirectories including hidden ones


BlackHC:
`nbdev_build_docs` for one searches for notebooks using `glob('**/*.ipynb')` which also collects hidden directories.

I'm using a Jupyter-Lab extension (https://github.com/krassowski/jupyterlab-lsp) which stores additional backups in a `.virtual_documents` folder, which lead to issues for me as they get picked up by `nbdev_build_docs`.

The current code in `nbdev_build_docs` ignores `.ipynb_checkpoints` specifically. Could this be expanded to ignore hidden directories in general?

Thanks,
 Andreas

hamelsmu:
Are you using the latest version of nbdev?  

`nbdev_build_docs` filters out hidden directories with this line of code:

```py
    if fname is None:
        files = [f for f in Config().path("nbs_path").glob('**/*.ipynb')
                 if not f.name.startswith('_') and not '/.' in f.as_posix()]
```

I tested it and it does ignore hidden directories.   You can upgrade to the latest nbdev by cloning this repo and running pip install -e .   Alternatively I believe that `pip install -U nbdev` should work because this was a change made 4 months ago.

BlackHC:
Thank you! I have the latest version installed. Please see below for links to code that has issues.

The following line causes issues with `nbdev_build_docs` (in master): 

https://github.com/fastai/nbdev/blob/7fcc358f9a4ab7709d0d5808bfb02a7456930eac/nbdev/export2html.py#L713

I've seen a similar issue in:

* `nbdev_read_nbs`: https://github.com/fastai/nbdev/blob/7fcc358f9a4ab7709d0d5808bfb02a7456930eac/nbdev/test.py#L130

* `nbdev_clean_nbs`: https://github.com/fastai/nbdev/blob/6c4934a2ddfa40ea27b5f229f21ee77aea1a479e/nbdev/clean.py#L83

* `nbdev_trust_nbs`: https://github.com/fastai/nbdev/blob/05b7a910044b3858217f240d6225c9a9e4f8a49d/nbdev/sync.py#L162

I hope this helps.

Thanks,
 Andreas

hamelsmu:
cc: @jph00 looks like potentially we are trying to find notebooks using slightly different or repeated logic in different places.  Might be worth thinking about this, or also at least keeping this in mind for v2

muellerzr:
[...]

PR #359, opened 2020-12-30
add ability to find notebooks recursively

### LINKED ISSUE #358: Feature Request: Allow for recursive modules
I've come to an interesting situation where some of my nbdev notebooks are nested, such as:

```
- nbs
  a.ipynb
  - nbs2
    b.ipynb
```
And both export modules. It would be nice to have nbdev work with that ( @hamelsmu )

hamelsmu:
Closes #358 

Note: I made sure that `.ipynb_checkpoints` is ignored by ignoring all paths that contain the substring `/.`, which ignores all hidden directories.

Ready for review: @jph00 

This also partially related to https://github.com/fastai/nbdev/issues/357.


jph00:
This should be optional and default to false, since otherwise it's a breaking change

hamelsmu:
ok @jph00 ready for review.  I made the following changes

- made recursive search optional
- updated the nbdev_build_lib CLI interface (I forgot to do that previously for some reason)
- moved the notebook glob search to its own function 
- Updated tests in `showdoc` that incidentally use this function to test something else about showing the signature

cc: @muellerzr

hamelsmu:
@jph00 made changes you requested

Issue #355, opened 2020-12-29
nbdev_nb2md doesn't work in Colab due to CR/CRLF


IRailean:
When converting from .ipynb to .md cells with #hide are present in the markdown file.

Notebook:

![Untitled](https://user-images.githubusercontent.com/38012331/103283123-5efbe300-49e0-11eb-9846-0b24b9f7b770.png)

Markdown file contains:

# Title

```
#print("hello")
```

Test1

```
#hide 
print("test cell")
```

    test cell

Test2

hamelsmu:
I'm not able to reproduce this behavior.  Steps I took:

1. Cloned `fastai/nbdev` repo
2. Made sure I am using the latest version of nbdev `pip install -e .`
3. navigated to notebook folder `cd nbs`
4. ran the following command `nbdev_nb2md 06_cli.ipynb`
5. when running `tail 06_cli.md` the last cell of the corresponding notebook is infact hidden.  

Closing this issue, but please do open a new issue that has a reproducible example with a real project & notebook that we can test against.  

cc: @jph00 

Thanks,

IRailean:
@hamelsmu Thank you for the clarification. It works with Jupyter Notebooks indeed. The problem was that I`ve been using Colab Notebook. As I know in Colab nb there is additional metadata which makes it different from Jupyter nbs. Is nb2md supposed to work with Colab nbs?

IRailean:
Colab adds to the end of some lines \r character. 
I created two identical notebooks, one in Jupyter Notebook, another in Colab. If you open them in text editor you will see, that in Jupyter it will look like:
'#hide\n'
and .ipynb from Colab is:
'#hide\r\n'

If you remove '\r' character and run nb2md all works ok.

jph00:
@IRailean can you please provide a step-by-step way to reliably reproduce this issue please?

IRailean:
Hi @jph00 . 
Here is how I reproduced the problem:
A.
1.Create a new notebook in Google Colab
2. Add 
```python
#hide
print("Hello world")
```
and run it
3. Create second separate Google Colab nb and run:
```python
!git clone https://github.com/fastai/nbdev.git
!pip install -e nbdev
```
to install nbdev
4. Run:
```python
import json
nb_path = '/content/drive/MyDrive/Colab Notebooks/Untitled7.ipynb'
with open(nb_path) as f:
  data = json.load(f)

print(data)
```
to get first notebook in json format (Untitled7 is the name of my first nb).
[...]

PR #364, opened 2021-01-04
Small fixes to have nbdev work better with Google Colab notebooks

muellerzr:
Adjusts metadata to see if `tags` is present, if not pop
Adjusts small issue in `source` where it can be an array of a blank string
Adjusts another issue in source where a random `\r` would be placed (fixes https://github.com/fastai/nbdev/issues/355)
Added some tests

Also now the first properly cleaned nbdev straight from google colab!

Fixes https://github.com/fastai/nbdev/issues/363

cc: @jph00


muellerzr:
Also going to try and handle this issue as well: https://github.com/fastai/nbdev/issues/355

TL;DR if you notice colab adds some /r randomly

muellerzr:
Okay @jph00 this is ready for review, let me know your thoughts please

muellerzr:
Converting to a draft for the time being. I think I can make this better. It doesn't account for if metadata is also in `outputs`, which apparently can happen.

nbdev/clean.py
```diff
@@ -43,6 +43,12 @@ def clean_cell(cell, clear_all=False):
     if 'outputs' in cell:
         if clear_all: cell['outputs'] = []
         else:         clean_cell_output(cell)
+    if cell['metadata'].get('tags') is not None: cell['metadata'].pop('tags')
```
jph00, line 46: `pop('tags',None)` would let you remove the `if` IIUC

nbs/07_clean.ipynb
```diff
@@ -168,6 +174,87 @@
     "               'source': 'awesome_code'})"
    ]
   },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "tst2 = {'cell_type': 'code',\n",
```
jph00, line 183: I'd suggesting simplifying this down to just the absolute minimum needed to test. Having lots of code in tests makes them hard to read and understand, and harder to maintain.

muellerzr: By moving this into `nbdev_clean_nbs` directly this test is no longer needed

muellerzr:
@jph00 summary of changes:

`clean_cell_output` checks for `tags` in the metadata. Since we don't actually want tags anyways I left it's default to `o`

Since we also just pop off the `tags` anyways I just got rid of it in `cell_metadata_keep`

nbs/07_clean.ipynb
```diff
@@ -168,6 +170,42 @@
...
+    "       'execution_count': 26,\n",
+    "       'metadata': {'tags':[]},\n",
+    "       'outputs': [{'execution_count': 2, \n",
+    "                    'data': {\n",
+    "                        'application/vnd.google.colaboratory.intrinsic+json': {\n",
```
jph00, line 184: There's still a lot of data here in the test. Can you minimize it to what's needed to test the logic?

jph00 (changes requested):

nbs/07_clean.ipynb
muellerzr: Let' me know if that's a better test

jph00:
Thanks!

Issue #368, opened 2021-01-08
nbdev_diff_nbs doesn't catch recursive


muellerzr:
Discussed with @hamelsmu, approach I'll be looking into is adding a flag for `settings.ini` for `recursive`, which can then be passed to `nbdev_build_lib` and `nbdev_diff_nbs`.

PR #370, opened 2021-01-08
Allow for recursive flag in settings.ini and adjust functions to use nbglob when searching for notebooks

muellerzr:
Closes https://github.com/fastai/nbdev/issues/368

Adjusts recursive to be a flag rather than an option so it works well with the tests and workflow. Default is `False`


muellerzr:
Note: two tests in `show_doc` needed to be changed due to `recursive` now being `None` by default in `notebook2script`

muellerzr:
@jph00 this is good to go now

PR #392, opened 2021-01-20
support windows

### LINKED ISSUE #393: 01_sync and 03_export2html failed on Windows
1. Unix Path isn't supported on Windows.
2. Parallel in notebook isn't supported on Windows

mszhanyi:
fix #393
1. 01_sync.ipynb and 03_export2html.ipynb failed in windows
2. The reasons are
   2.1 Unix path is not applicable on Windows
   2.2 Notebook could not use parallel on Windows

image


nbdev/export2html.py
```diff
@@ -553,7 +553,11 @@ def notebook2html(fname=None, force_all=False, n_workers=None, cls=HTMLExporter,
                 files.append(fname)
     if len(files)==0: print("No notebooks were modified")
     else:
-        passed = parallel(_notebook2html, files, n_workers=n_workers, cls=cls,
+        if sys.platform == "win32":
```
jph00, line 556: I believe you should be able to just set `n_workers=0` for win32. fastcore's `parallel` disables multiprocessing in that case.

jph00 (changes requested):
Many thanks for the PR! Just one suggestion to try out please.

nbdev/export2html.py
mszhanyi: @jph00 , Thank your reply. I'm afraid that we could NOT use n_workers=0 here since it still spawns a new process even n_wordkers=0

we could verify it by the below snippt.
```
from concurrent.futures import ProcessPoolExecutor as Executor

def func():
        print(__name__)
        print('Hello')

if __name__ == '__main__':
        print(__name__)
        with Executor(1) as executor:
                future=executor.submit(func)
                print(future.result())
```

we can see the result is
__main__
__mp_main__
Hello
None

ref:
https://stackoverflow.com/questions/61860800/running-a-processpoolexecutor-in-ipython

mszhanyi: @jph00 , Thank your reply. it was stuck when if n_workers=0. 
I find it's caused by https://github.com/fastai/fastcore/blob/8ddf4dde118dea53daf849d7a5f2dc425ca49390/fastcore/parallel.py#L84-L85.
The Manager().Lock() spawns a new process.

Could we don't set [self.lock = Manager().Lock()](url) when it is not_parallel?

jph00:
`parallel` uses fastai's ProcessPoolExecutor, not the stdlib one, which behaves differently with n_workers=0. Try it :)

jph00:
> @jph00 <https://github.com/jph00> , Thank your reply. it was stuck when if n_workers=0.
> I find it's caused by https://github.com/fastai/fastcore/blob/8ddf4dde118dea53daf849d7a5f2dc425ca49390/fastcore/parallel.py#L84-L85.
> The Manager().Lock() spawns a new process.

> Could we don't set self.lock = Manager().Lock() <https://www.fastmail.com/mail/url?u=35e541ae> when it is not_parallel?


Ah great find - yes feel free to submit a PR to add a condition there. (And please at-mention me on any PRs you do)

mszhanyi:
git diff throws exception without https://github.com/mszhanyi/nbdev/blob/93ceef10b29e2d897f2e14d8ed1c81fbcab0bc53/nbdev/clean.py#L85

jph00:
Super - thank you :)

PR #412, opened 2021-02-02
extend test timeout for windows

mszhanyi:
Since we have to use single process in notebook on Windows, some operations take longer time.
For example , the below takes 10 more minutes.
https://github.com/fastai/fastai/blob/master/nbs/14_callback.schedule.ipynb
image

### DIFF
--- a/nbdev/test.py
+++ b/nbdev/test.py
@@ -79,7 +79,8 @@ def test_nb(fn, flags=None):
         nb = read_nb(fn)
         for f in get_all_flags(nb['cells']):
             if f not in flags: return
-        ep = NoExportPreprocessor(flags, timeout=600, kernel_name='python3')
+        timeout = 1800 if sys.platform == "win32" else 600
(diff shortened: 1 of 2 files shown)


mszhanyi:
@jph00  fastai/fastai#3198 depends on this PR

jph00:
Could you please add a `#slow` comment to the top of any cell that takes too long in Windows, or see if you can change it to run more quickly? (e.g. the above example you provide looks to be just an integration test, so it's fine to be marked slow - or alternatively a smaller dataset could be used)

mszhanyi:
More info:
1. Linux 12 core 2.2GHz CPU, 2 1080Ti, 53 seconds to complete 14_callback.schdule.ipynb
2. Windows 4 K80, CPU 2.6GHz, 804 seconds to complete 14_callback.schdule.ipynb

mszhanyi:
reduce time by adding #slow

Issue #381, opened 2021-01-16
nbdev_nb2md throws error when called in a notebook


urspb:
When I tried to call `nbdev_nb2md` I got the following error, which I don't know how to handle:
![image](https://user-images.githubusercontent.com/50357297/104809836-7c0e1f80-57f0-11eb-88fb-8ccd54089ac5.png)

I'm Using nbdev 1.1.11

    ➜  nb2ltx git:(master) ✗ pip list | grep nbdev
    nbdev              1.1.11

An other issue appeared, when I tried to call `nbdev_nb2md` on the command line. If I try to use the argument  `--img_path` I get an error:

    ➜  nb2ltx git:(master) ✗ nbdev_nb2md --dest tex --img_path 'tex/images' 10_Code.ipynb
    usage: nbdev_nb2md [-h] [--dest DEST] [--use_img USE_IMG] [--replace REPLACE] path_nb
    nbdev_nb2md: error: unrecognized arguments: --img_path 10_Code.ipynb

What puzzles me is the `[--use_img USE_IMG]` in the usage section of the error message. There seems to be a connection to the error message from the notebook because `--use_img` is mentioned there also . If I execute `nbdev_nb2md  -h` on the command line there is no option `--use_img` mentioned.

    ➜  nb2ltx git:(master) ✗ nbdev_nb2md  -h
    usage: nbdev_nb2md [-h] [--dest DEST] [--img_path IMG_PATH] [--jekyll JEKYLL] fname
    
    Convert the notebook in `fname` to a markdown file
    
    positional arguments:
      fname                A notebook file name to convert
    
    optional arguments:
      -h, --help           show this help message and exit
      --dest DEST          The destination folder (default: .)
      --img_path IMG_PATH  Folder to export images to (default: )
      --jekyll JEKYLL      To use jekyll metadata for your markdown file or not (default: False)

Having read the contribution guidelines, I tried to call `import fastai.utils.collect_env; fastai.utils.collect_env.show_install(1)` but was stopped by an error message. Is `fastai.utils.collect_env` still part of the current fastai version?

hamelsmu:
This works for me 

![image](https://user-images.githubusercontent.com/1483922/105257662-1944bd00-5b3d-11eb-9038-b4da2392f241.png)

also I tried on a seperate folder/repo and works too.  Can you try nbdev_nb2md on a simple notebook with just print('hello') in it?    

[...]

PR #424, opened 2021-02-16
fix nb2md

hamelsmu:
A `@call_parse` function cannot currently call another `@call_parse` function. The function's inputs end up getting mangled when there are nested call parse functions.  This is what is causing the issue reported in #381 

This fixes that by reorganizing the relevant code such that `@callparse` is only a wrapper around the functionality we want for this particular piece of code.   cc: @jph00 

I have tested this to make sure it resolves the issue.


hamelsmu:
Another way to fix this would be to fix `@call_parse` somehow.  I had a go but got stuck on that - but fixing fastcore's callparse is another option.  @jph00 WDYT?

PR #449, opened 2021-03-30
minor clarifications in CONTRIBUTING and tutorial

drscotthawley:


drscotthawley:
As per [this forum thread](https://forums.fast.ai/t/nbdev-tutorial-gh-pages-build-fails-missing-expected-setting-description/87022/31) with @hamelsmu

PR #488, opened 2021-05-11
Docs nav doesn't work on gitlab

tcapelle:
Removes .html names from sidebar.
Fixes html 404 on gitlab hosting
you can check here:
https://thomas.capelle.gitlab.io/nbdev_template/


jph00:
Many thanks.

Issue #505, opened 2021-07-03
notebook2script fails on notebooks saved by VS Code native editor


sreiz1:
Since the recent VS Code update I've started using the native notebook editor with VS Code keybindings etc. It's great, but apparently changes the notebook saved format a bit (I noticed that metadata for the notebook is now saved at the end).

This appears to confuse nbdev notebook2script - it still runs without errors, but the created python files are mostly empty, skipping most or all #export cells. (notebook2script worked fine before with the old jupyter notebook editor on the same notebooks).

This is with fastai 2.4 and nbdev 1.1.14. NB The instructions in the contributing guidelines for show_install don't work anymore.

HDavidEspinosa:
I am having the same issue. The worst part is if one installs the new Pylance release on VSCode, it updates the notebook editor and breaks the script. 

As far as I can tell, the new native editor changes the cell data of notebooks.

I noticed that the cell structure used to be:
"cells": [
  {
   "cell_type": "code",
   "execution_count": null,
   "metadata": {},
   "outputs": [],
   "source": [
    "# default_exp rpl_summarise"
   ]
  },

But after editing the cell with VSCode's new notebook editor, it is:

 {
   "cell_type": "code",
   "execution_count": 1,
   "source": [
    "# default_exp rpl_summarise"
   ],
   "outputs": [],
   "metadata": {}
  },

Note that "outputs" and "metadata" are now after "source", not before.  ALso, their order is now inverted. 

I re-runned the script after manually reorganising the elements, and it seems to correct the problem in the case of the file's first, cell. But the same solution did not work with a cell in the middle of the document, past other cells that have been edited.

HDavidEspinosa:
Here is a nbdev project that I created using VSCode. As anyone can see, it does not build the core module. 

This might help reproduce the problem. I looked at the 00_export.ipynb of nbdev as hard as I could, but I couldn't hack it.

I have the filling that something in the new notebook structure may be braking the regex generated by `_mk_flag_re` but I am not sure.
[vsc_test.zip](https://github.com/fastai/nbdev/files/6854766/vsc_test.zip)

jph00:
Should be fixed in v2

PR #513, opened 2021-07-22
Patch for VSCode problem 

HDavidEspinosa:
Hello

In response to the following Issue report: [notebook2script fails on notebooks saved by VS Code native editor #505](https://github.com/fastai/nbdev/issues/505), I have checked the 00_export notebook to try to patch the problem. 

As noted in the commit history, I think the solution is:
Added lines 6 and 7 to check_re function to avoid vscode transformation problem.
Added test after notebook2script cell. The test seems to indicate that the vscode
problem should be fixed.

Since making the changes in this branch I have managed to work with vscode again. 

As noted in the following [comment](https://github.com/fastai/nbdev/issues/505#issuecomment-884068801), the problem is that the new vscode editor adds \r\n characters to new lines. At least in windows. 

The patch, makes sure that all \r\n are simple \n characters before applying the regex.

I also added a test to check this issue has been solved. However, I suspect that github might replace the characters on push. In any case, I uploaded the test notebook in a zip in the afford mentioned issue.

I hope this helps others I love nbdev and hope it sticks around for a long while.


jph00:
Many thanks @HDavidEspinosa . I'm not able to merge this directly, since it's got a lot of unrelated changes across 21 files. Instead I'll just make the edit directly, if that's OK.

Many thanks for looking into this issue.

HDavidEspinosa:
Sounds great cheers.

HDavidEspinosa:
@jph00 sorry to keep bugging with this issue.

Today I reviewed the solution and found that, though it works, it also introduces a problem. In notebooks with \r\n the built modules end up with double  \n which can break some modules. 

The better solution is to make two changes to the export module. 
in the check_re function  the last three lines should be:
    cell_source = cell['source'].replace('\r', '') # Eliminates the \r\n bug
    result = pat.search(cell_source)
    return result

The second change is to the split_flags_and_code function, there the first line of the function can be replaced for:
    source_str = cell['source'].replace('\r', '')
    code_lines = source_str.split('\n')
instead of 
    code_lines = cell['source'].replace('\n')

This works fine. Thanks again

jph00:
@HDavidEspinosa please don't apologize for helping! :) Thanks for your suggestion - I've added that now.

Issue #517, opened 2021-08-12
nbdev testing documentation


cedarmora:
I had a pretty hard time figuring out that the documentation for the [nbdev tests](https://nbdev.fast.ai/test.html) (e.g. `test_eq`) is actually in [fastcore](https://fastcore.fast.ai/test.html), could there be some link or some bold sentence near the top of the [nbdev test documentation](https://nbdev.fast.ai/test.html#get_all_flags) explaining this?

yegeniy:
@cedarmora - see the tip in https://nbdev.fast.ai/example.html#Where-The-Magic-Happens:-Write-Documentation-And-Tests and https://github.com/fastai/nbdev/blob/master/CONTRIBUTING.md

That might be a good way to get started.

<img width="851" alt="image" src="https://user-images.githubusercontent.com/54481302/129285995-3708b132-0c05-4a08-bed3-72acc1e25b4e.png">

cedarmora:
Ok I see those two, thank you for pointing that out. 

But as somebody who is just learning nbdev, that's not where I would have expected to learn about the tests, I would have expected to find this information on [this page](https://nbdev.fast.ai/test.html) since it's the only tab in the nbdev documentation that has the word "tests" in it.

I'll see if I can whip up a PR

yegeniy:
Good luck. I'm just learning this system as well.

I've found https://nbdev.fast.ai/nbdev_comments.html to be a great
reference, once you're really diving in beyond the tutorials.

You may find a typo or two there as well. 😉

yegeniy:
I see the source of your confusion, but just to help you clarify it:

https://nbdev.fast.ai/test.html is actually part of the documentation for the source code of nbdev itself. 

Specifically, it documents the module called `test`:
- https://github.com/fastai/nbdev/blob/master/nbs/04_test.ipynb (the `test` notebook)
- https://github.com/fastai/nbdev/blob/master/nbdev/test.py (and the generated code for `test`) 

As the documentation there says, the `test` module is there to let `nbdev` "extract tests".

<img width="1096" alt="image" src="https://user-images.githubusercontent.com/292802/129292504-615805df-f5b1-41d4-b49d-df89424de2e7.png">

cedarmora:
Ok I made a PR, let me know if it needs any changes.

PR #518, opened 2021-08-13
Add reference to the fastcore tests in the Extract tests tab

cedarmora:
Fixes issue #517

As somebody who is just learning nbdev, I was trying to figure out where nbdev testing functions come from, specifically `test_eq()`. So I went to the nbdev docs and clicked the first tab that has **test**,  on it and to my surprise did not find the `test_eq` function definition. After some exploration I discovered the tests are actually defined in [fastcore](https://fastcore.fast.ai/test.html), but this was not obvious. This PR adds a link on the nbdev testing page to the fastcore testing page to make this relationship more obvious.

**Before:**
before

**After:**
after


cedarmora:
Ok nevermind then.

PR #527, opened 2021-09-08
Update path to _config.yml in tutorial

mjavorka:
Current version of tutorial shows path to `_config.yml` file in root directory. The config file is stored in the `docs` directory.

This PR fixes the path to the file in the tutorial.


hamelsmu (approved):

hamelsmu (changes requested):
You have to change the notebook, not just the HTML :)  this project is written with nbdev, so the best route is to change the notebook that generates this page!

mjavorka:
> You have to change the notebook, not just the HTML :) this project is written with nbdev, so the best route is to change the notebook that generates this page!

Thank you for the review. In this MR I update the notebook (`.ipynb` file), not the generated HTML. But I created new PR https://github.com/fastai/nbdev/pull/531, where both the notebook and HTML are updated.

PR #543, opened 2021-10-26
Fix nested nb sync

danielmitchell:
This makes a small change to _script2notebook to enable syncing changes back to a notebook in a nested folder structure, e.g. nbs/subdir/nb.ipynb. The current behavior skips them silently and only syncs to nbs in the root dir.

There were no tests so I created a set that verifies the normal non-nested case as well as this change.

The nested non-default export test case would fail if using the current export syntax, but passes when the nested subdir is prefixed to the _Comes from nb.ipynb_ comment. I plan to submit a separate PR to update 00_export.ipynb to prefix any subdirs for a non-default export. The test case I included here reflects that change since this function already supports it.


jph00:
Many thanks!

jph00:
Sorry I'm not sure how I missed it, but it turns out this PR broke CI on Windows. I've had to revert it as a result. Feel free to resubmit if you're able to get a cross-platform version working.

PR #571, opened 2022-01-09
Minor typo fix: pyipi -> pypi

tylere:


muellerzr:
Thanks for this! cc: @hamelsmu or @jph00

PR #573, opened 2022-01-14
Make relative imports configurable

anubhavmaity:
Using nbdev for my personal projects, I prefer nbdev will convert the notebooks to python scripts with absolute imports instead of relative. It really helps me to run with some other frameworks also.
(Will squash and merge)


anubhavmaity:
@muellerzr
Can you please review this PR?

muellerzr:
This would be up to @jph00 for if we want to include it.

jph00:
Thanks for the PR, but I'd rather not add this functionality.

PR #575, opened 2022-01-23
Add docments to nbdev!

muellerzr:
# What does this do?

This PR with the help of KevinB on the forums helps bring `docments` directly into nbdev! 
It's formatted in such a way where if they don't use docments, you still get the added benefits of typings in a pretty table showing in `show_doc` and `doc`!

There was some custom CSS involved to handle peculiar situations with Jupyter where it would not output a table in the popout viewer, so we had to inject some in there.

## Example Renderings
Original function:
```python
def addition(
    a:int, # first number to add
    b:float, # second number to add
    # Sum of `a` and `b`
) -> float:
    "Adds two numbers together"
    return a+b
```
Rendering with `show_doc`:
image

Popout with `doc`:
image

When no docments are involved:
```python
def addition(a:int,b:int) -> float:
    "Adds `a` and `b`"
    return a+b
```
Rendering with `show_doc`:
image

With `doc`:
image

## Added tests
Most of the tests revolved around making sure the helpers for formatting the return annotations and input annotations from docments were how we wanted them to be done. 

cc: @jph00 and @hamelsmu :)


nbdev/showdoc.py
```diff
@@ -243,9 +244,64 @@ def _format_cls_doc(cls, full_name):
...
+def _format_args(elt):
+    "Generates a formatted argument string"
+    ment_dict = docments(elt, full=True)
+    ret = None
+    arg_string = ""
+    return_string = ""
+    if "self" in ment_dict.keys(): ment_dict.pop("self")
+    if "cls" in ment_dict.keys(): ment_dict.pop("cls")
+    if "return" in ment_dict.keys():
+        ret = ment_dict["return"]
```
jph00, line 294:
```suggestion
        ret = ment_dict.pop("return") # and remove next line
```

nbdev/showdoc.py
```diff
@@ -243,9 +244,64 @@ def _format_cls_doc(cls, full_name):
...
+        ment_dict.pop("return")
+    if len(ment_dict.keys()) > 0:
```
jph00, line 296: Lots of these single line bodies can be put on previous line

nbdev/showdoc.py
```diff
@@ -243,9 +244,64 @@ def _format_cls_doc(cls, full_name):
...
+    for key, item in argument_dict.items():
+        is_required=True
+        if key == 'return': continue
+        if item['default'] != inspect._empty:
+            is_required = False
+        arg_string += f"|**`{key}`**|"
+        arg_string += "*None Specified*|" if item['anno'] == inspect._empty else f"`{item['anno']}`|"
+        arg_string += "*No Default*|" if is_required else f"`{str(item['default'])}`|"
+        if _has_docment:
+            if item['docment'] is not None:
```
jph00, line 266: Use ternary operator for this

nbdev/showdoc.py
```diff
@@ -243,9 +244,64 @@ def _format_cls_doc(cls, full_name):
...
+    arg_string = ""
+    return_string = ""
+    if "self" in ment_dict.keys(): ment_dict.pop("self")
+    if "cls" in ment_dict.keys(): ment_dict.pop("cls")
+    if "return" in ment_dict.keys():
+        ret = ment_dict["return"]
+        ment_dict.pop("return")
+    if len(ment_dict.keys()) > 0:
+        arg_string = _generate_arg_string(ment_dict)
+    if not ret["anno"] == inspect._empty:
```
jph00, line 298: might need `.get` with a default here, in case `anno` doesn't exist?

nbdev/showdoc.py
```diff
@@ -243,9 +244,64 @@ def _format_cls_doc(cls, full_name):
...
+# Cell
+def _format_args(elt):
+    "Generates a formatted argument string"
+    ment_dict = docments(elt, full=True)
+    ret = None
+    arg_string = ""
+    return_string = ""
+    if "self" in ment_dict.keys(): ment_dict.pop("self")
+    if "cls" in ment_dict.keys(): ment_dict.pop("cls")
+    if "return" in ment_dict.keys():
```
jph00, line 293: Use `{}` as a default param to `pop` so can remove this conditional?

nbdev/showdoc.py
```diff
@@ -243,9 +244,64 @@ def _format_cls_doc(cls, full_name):
...
+    return return_string
+
+# Cell
+def _format_args(elt):
+    "Generates a formatted argument string"
+    ment_dict = docments(elt, full=True)
+    ret = None
+    arg_string = ""
+    return_string = ""
+    if "self" in ment_dict.keys(): ment_dict.pop("self")
```
jph00, line 291: None of these conditionals are needed - just add the optional `default` param to `pop`

nbdev/showdoc.py
```diff
@@ -258,14 +314,20 @@ def show_doc(elt, doc_string=True, name=None, title_level=None, disp=True, defau
...
-        try: monospace = (get_config().get('monospace_docstrings') == 'True')
+        try: monospace = (Config().get('monospace_docstrings') == 'True')
         except: monospace = False
         # doc links don't work inside markdown pre/code blocks
         s = f'```\n{s}\n```' if monospace else add_doc_links(s, elt)
         doc += s
+    if len(args) > 0 and "arg" not in s.lower():
+        try:
+            doc += f"\n\n{_format_args(elt)}"
+        except:
```
jph00, line 329: This `except` clause should be more limited, and documented. Even better if you can avoid it

jph00 (changes requested):
Many thanks for this. Can you describe more how you decide whether there are docments, or whether there is just some general comment in the code?

In the types in the table, can you show just the type name, without the `class` prefix etc?

muellerzr:
I don't actually check for that, I just run `docments` natively. I do include an option of if arguments are already documented in the string, we ignore `docments` and just render it that way. 

If you have ideas on checking for that I'd be open to it.

nbdev/showdoc.py
muellerzr: docments will always return an `anno` in the dictionary, even if there is none. So I don't think it's needed.

muellerzr:
re: tables, used `qual_name` to get us there

nbdev/showdoc.py
muellerzr: Now that it's been iterated over more, this isn't really needed. It was there as a catch-all in case `docments` failed, but docments really *shouldn't* fail given its nature. 

What we may want to do instead (and this goes with the displaying/figuring out if it's docments or not) is to include `show_docments` as a boolean at the end of `show_doc` and `doc`, that is set to `True` by default. WDYT?

muellerzr: Went with that route as well as an explicit warning:

```python
    if len(args) > 0 and show_docments:
        try:
            doc += f"\n\n{_format_args(elt)}"
        except AttributeError:
            print(f"Warning: `docments` annotations are unavailable for {qual_name(elt)} and will not be shown.")
            pass
```
The issue with this is something about the `DocsTestClass` it will not let `docments` grab or do anything, it raises an issue. Not 100% sure why that's the case, but opt'd for just an explicit warning

kevinbird15:
You probably want to also force a rebuild of all the docs so that nbdev uses docments everywhere, not just where the docs have explicitly changed. `nbdev_build_docs --force_all True`

kevinbird15:
Found one where it doesn't seem to be quite right: 

image
coming from nbdev's export2html.html doc page

kevinbird15:
This one definitely isn't correct currently: 

https://github.com/fastai/nbdev/blob/730f804cdb9a6827622ae37b9e514d8e0d1f099a/nbdev/export2html.py#L457-L462

image

kevinbird15:
```
import pandas as pd
show_doc(pd.DataFrame)
```

Here is the signature: 
```
pd.DataFrame(
    data=None,
    index: 'Axes | None' = None,
    columns: 'Axes | None' = None,
    dtype: 'Dtype | None' = None,
    copy: 'bool | None' = None,
)
```
Here's how it is rendering (dang pipe...): 
image

nbdev/showdoc.py
kevinbird15: Found an example where this happens in case it's helpful: 
```
import numpy as np
show_doc(np.arange)
```

muellerzr: I don't actually see where that's ever defined, so it may be a meta programming issue? I think just having the `doesn't support XYZ for docments` is fine as something to revert to, and just show what is normally there.

@jph00 I think what we should do in this case is have monospace be toggleable so it renders well. eg:
```python
show_doc(np.arange, show_docments=False, monospace=True)
```
is how we "fix" the numpy issue. Basically if folks want to keep using that type of documentation string, we don't markdown format it

muellerzr:
Here's our issue with `nbdev_build_lib`:

Currently `docments` doesn't know what a `Param` is, and how it should be read, because doing `docments` for `nbdev_build_lib` shows us that `bare`'s `anno` is the type and the annotation:
```python
{'fname': - docment: None
 - anno: str <A notebook name or glob to convert>
 - default: None,
 'bare': - docment: None
 - anno: <Omit nbdev annotation comments (may break some functionality).>
 - default: False,
 'return': - docment: None
 - anno: <class 'inspect._empty'>
 - default: <class 'inspect._empty'>}
```
So would we consider this a bug in `docments` currently @jph00 that should get solved?

muellerzr:
Pipe issue is solved by adding an `or` to the docstring. That's the only situation it becomes a problem simply because the table is also a pipe, so it was adding an extra cell

muellerzr:
Took me a bit but I solved the issue with anything coming from typing. TL;DR is adding this little function:
```python
#export
def _format_annos(anno):
    "Returns a clean string representation of `anno` from either the `__qualname__` if it is a base class, or `str()` if not"
    annos = listify(anno)
    new_anno = "(" if len(annos) > 1 else ""
    def _inner(o): return o.__qualname__ if type(o) == type else str(o)
    for i, anno in enumerate(annos):
        new_anno += _inner(anno)
        if len(annos) > 1 and i < len(annos) - 1:
            new_anno += ', '
    return f'{new_anno})' if len(annos) > 1 else new_anno
```
It's goal is to return a string that is our clean and pretty representation. Here's some tests for you to look at @kevinbird15 to see what it does:
```python
#hide
from typing import Union, Tuple
test_eq(_format_annos(Union[int,float]), 'typing.Union[int, float]')
test_eq(_format_annos(Tuple[int,float]), 'typing.Tuple[int, float]')
test_eq(_format_annos((int,float)), '(int, float)')
test_eq(_format_annos(int), 'int')
```

muellerzr:
Finally got to a point where our script is stable.

To run make sure your directory is as such:
- `script.py`
- `nbdev_library_git_repository`

If it's not an nbdev library, it will just try and convert all .py and .ipynb files available. Otherwise it'll use the path. 

Other important key is we flag if the user has any logic with the `Param` class, and warn them about how they should remove it, and the file locations of them.

https://gist.github.com/muellerzr/ab5756a40a3ad5fc59331996e0776b7f

kevinbird15:
```
import torch
show_doc(torch.nn.L1Loss)
```

It looks like None in the Return is returning as `` now.  I'm guessing that should be a blank return or maybe that happens if no return type is specified? I'm not exactly sure what `-> None` tells us in this case.

```
Init signature: torch.nn.L1Loss(size_average=None, reduce=None, reduction: str = 'mean') -> None
```
image

kevinbird15:
Possibly fixed with this PR to muellerzr:docment-integration: https://github.com/muellerzr/nbdev/pull/1

kevinbird15:
Here's another one: 
https://github.com/fastai/nbdev/blob/48b9acf6d6a96d211d0fe43ce9cb84bbc81613da/nbdev/export.py#L407-L416

image

I will look at this tomorrow if nobody beats me to it

nbdev/showdoc.py
muellerzr: Solution fixed. The issue was builtin modules cannot have source, so `getsource()` failed. Put in an explicit print about it.

muellerzr:
@jph00 this is ready for another passthrough!

muellerzr:
For the curious mind, `def _inner(o): return o.__qualname__ if '<' in str(o) else str(o)` is needed because in Python 3.6, doing `__qualname__` with the Typing library still returns just the base class type (like `Union`) rather than `Union[a,b]`. Otherwise the old logic would hold true.

jph00:
Isn't py36 unsupported now though?

muellerzr:
@jph00 it did. I can disregard that and put in a better choice then. We should also update the CI/CD

muellerzr:
@jph00 I can shortly after this is integrated. I'll update this one though.

Also added a setting in the config for showing all docments + tested on a few libraries we converted on and looks good. We should expect failing on py 3.6

Key is called `show_all_docments` and by default it is `False`

muellerzr:
Wondering if something like
`force_docments` or `docments_everywhere` might be better

kevinbird15:
Currently seeing this for everything: 
```
Warning: `docments` annotations will not work for built-in modules, classes, functions, and `enums` and are unavailable for HTMLParser.__init__. They will not be shown
```

Coming out of this statement: 
```
        if elt.__class__.__module__.replace('_','') == 'builtins' or is_enum(elt):
            if not silent:
                print(f'Warning: `docments` annotations will not work for built-in modules, classes, functions, and `enums` and are unavailable for {qual_name(elt)}. They will not be shown')

```

It looks like `elt.__class__.__module__` always is 'builtins' from what I can see.

Here is an example: 
```
from nbdev.cli import nbdev_clean_nbs
nbdev_clean_nbs.__class__.__module__
```
outputs
```
'builtins'
```

I think this should potentially be 
```
        if elt.__class__.__qualname__ == 'builtin_function_or_method' or is_enum(elt):
            if not silent:
                print(f'Warning: `docments` annotations will not work for built-in modules, classes, functions, and `enums` and are unavailable for {qual_name(elt)}. They will not be shown')

```

But this raises some other failed tests: 

```
show_doc(DocsTestClass)
```

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Input In [76], in <module>
      1 #hide
----> 2 show_doc(DocsTestClass)

Input In [73], in show_doc(elt, doc_string, name, title_level, disp, default_cls_level, show_all_docments, silent)
     31             print(f'Warning: `docments` annotations will not work for built-in modules, classes, functions, and `enums` and are unavailable for {qual_name(elt)}. They will not be shown')
     32     else:
---> 33         if _has_docment(elt) or show_all_docments:
     34             doc += f"\n\n{_format_args(elt)}"
     35 if disp: display(Markdown(doc))

Input In [49], in _has_docment(elt)
      2 def _has_docment(elt): 
----> 3     comments = {o.start[0]:_clean_comment(o.string) for o in _tokens(elt) if o.type==COMMENT}
      4     params = _param_locs(elt, returns=True)
      5     comments = [_get_comment(line,arg,comments,params) for line,arg in params.items()]

File /usr/local/lib/python3.8/dist-packages/fastcore/docments.py:24, in _tokens(s)
     22 "Tokenize Python code in string or function object `s`"
     23 if isfunction(s): s = getsource(s)
---> 24 return tokenize(BytesIO(s.encode('utf-8')).readline)

AttributeError: 'wrapper_descriptor' object has no attribute 'encode'
```

It looks like what is happening here is that `s = getsource(s)` doesn't get called because DocsTestClass is not a function so `s` is still the incoming `elt` from `_tokens`.

muellerzr:
@jph00 I am now thoroughly confident in this PR. I have pretty much checked every single possible situation we can be in where we extract the source code, and how we check for this is more inline for how we would grab our docments. Please let me know any changes you would like, but I would consider this finalized now!

muellerzr:
CI/CD fail might be from committing too soon, would recommend rerunning as all tests pass locally

jph00:
OK re-running CI now.

muellerzr:
For some reason post building the docs the CI failed. There wasn't a difference in what I did there so not sure what happened. gh actions is acting very odd... (I also can't see the log of errors)

muellerzr:
Esp weird considering when I enable actions on my fork, it's all blue 
image

jph00:
lgtm!

PR #578, opened 2022-01-31
docs fix collapse flags

### LINKED ISSUE #410: #collapse_input is not working
#collapse_input is not working.
example from docs:
https://nbdev.fast.ai/export2html.html#Collapsable-Code-Cells

I don't see any button or link to hide input.

ayasyrev:
To collapse input use #collapse, #collapse_show instead of #collapse_input, #collapse_input show.
Change information in docs.
#410


nbs/03_export2html.ipynb
```diff
@@ -65,7 +65,7 @@
     "- put `#hide_input` at the top of a cell if you don't want code to be shown in the docs\n",
```
kevinbird15, line 65: <p>So is this pr just updating the documentation to match what actually works to collapse or collapse_show a cell?  This isn't changing the wording from collapse_input</p><p> </p><p>So would it be fair to say that anywhere in the documentation we see <code>#collapse_input open</code> we would want to replace that with <code>#collapse_show</code> and anywhere in the documentation we see <code>#collapse_input</code>  after that find and replace, we would want to replace with just <code>#collapse</code>  or include another flag <code>('collapse_hide', '#collapse_input'),</code>  <span class="ql-cursor">﻿</span></p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/fastai/nbdev/pull/578/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='51'/>

kevinbird15: I also noticed that there are some changes to be made in this notebook as well: https://github.com/fastai/nbdev/blob/master/nbs/nbdev_comments.ipynb (happy to PR to your branch if you want help!)

kevinbird15: actually it looks like you already did this, I was looking at your master branch instead of docs_collapse!  I think this looks great from my perspective.

ayasyrev: I just remove from docs #collapse_input, because from code we need just #collapse. We can add to code parsing to it (just add it to _re_cell_to_collapse_closed), but simple solution correct doc.

jph00:
Many thanks - merged through CLI due to conflicts.

Issue #580, opened 2022-02-06
show_doc error in vscode notebook


danielmitchell:
vscode can run jupyter notebooks but running show_doc causes an error: `NameError: name 'Markdown' is not defined`.

This is related to [this issue](https://github.com/fastai/fastcore/issues/365) in fastcore where fastcore.utils.IN_NOTEBOOK returns False in vscode, even though in_notebook() returns True.

showdoc.py has a conditional import which doesn't run and causes the error.

```
if IN_NOTEBOOK:
    from IPython.display import Markdown,display
    from IPython.core import page
```

jph00:
I think this should work now

PR #581, opened 2022-02-06
Replaced IN_NOTEBOOK to fix vscode show_doc error

danielmitchell:
This PR fixes https://github.com/fastai/nbdev/issues/580

This is related to [this issue](https://github.com/fastai/fastcore/issues/365) in fastcore where fastcore.utils.IN_NOTEBOOK returns False in vscode, even though in_notebook() returns True.

Removing IN_NOTEBOOK and using a try/except ModuleNotFoundError block seems like a Pythonic way to handle this and fixes the error.

### DIFF
--- a/nbdev/showdoc.py
+++ b/nbdev/showdoc.py
@@ -9,11 +9,12 @@ from .imports import *
 from .export import *
 from .sync import *
 from nbconvert import HTMLExporter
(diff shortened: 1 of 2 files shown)


jph00:
Yup makes sense to me! I'll go ahead and fix this locally to avoid the conflict.

PR #582, opened 2022-02-08
Refactor try/catch to be better

muellerzr:
Refactored most cases where only `try: catch:` exists.

There were some pre the last PR that I couldn't quite figure out, such as failing to render `page.page`. For that, might want to mimic how the function itself raises errors I imagine? (Source [here](https://github.com/ipython/ipython/blob/0f4a73c91db71cde174275337fc5e226acc9dd7d/IPython/core/page.py#L177))

### DIFF
--- a/nbdev/showdoc.py
+++ b/nbdev/showdoc.py
@@ -362,7 +362,8 @@ def show_doc(elt, doc_string:bool=True, name=None, title_level=None, disp=True, 
     if doc_string and inspect.getdoc(elt):
         s = inspect.getdoc(elt)
         # show_doc is used by doc so should not rely on Config
-        try: monospace = (get_config().get('monospace_docstrings') == 'True')
+        # We use except because `get_config` raises Exception if not found
+        try: monospace = get_config().d.getboolean('monospace_docstrings', False)
         except: monospace = False
         # doc links don't work inside markdown pre/code blocks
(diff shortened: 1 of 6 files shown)


nbdev/showdoc.py
```diff
@@ -362,7 +362,8 @@ def show_doc(elt, doc_string:bool=True, name=None, title_level=None, disp=True,
     if doc_string and inspect.getdoc(elt):
         s = inspect.getdoc(elt)
         # show_doc is used by doc so should not rely on Config
-        try: monospace = (get_config().get('monospace_docstrings') == 'True')
+        # We use except because `get_config` raises Exception if not found
+        try: monospace = get_config().d.getboolean('monospace_docstrings', False)
         except: monospace = False
```
jph00, line 367: Can you constrain the `except` clause to just the type we need to catch, or remove the `try/catch` if we don't need it any more?

jph00 (changes requested):
Many thanks - just one request

nbdev/showdoc.py
muellerzr: @jph00 we can safely remove the try/catch, since `get_config` will go ahead and make its own config if one isn't found, and we include a default for `getboolean` if it's not found

muellerzr:
Good to go @jph00!

PR #585, opened 2022-02-17
Make sure docments have linking capability

muellerzr:


muellerzr:
@jph00 I can't add a label of enhancement to this, so you have to be the one to do so

Issue #587, opened 2022-02-21
`show_doc` types break when multiple are present


muellerzr:
Example:
![image](https://user-images.githubusercontent.com/7831895/154871365-0f804b58-2c5f-4f4f-984e-3869752cd128.png)

PR #588, opened 2022-02-21
Fix repr issue with `show_doc`

muellerzr:
Solves https://github.com/fastai/nbdev/issues/587 by having `type_repr` use `_format_annos` rather than `get_name`

image

The other option to consider is just completely change how `get_name` works and have it be `_format_annos.inner`

cc @jph00 

(Also mark this as Enhancement)


jph00:
Great!

PR #589, opened 2022-02-21
Add in repr for delegates

muellerzr:
Adds in support for repr with docments and `show_doc`. Example rendering below

image

@jph00 let me know thoughts on this one definitely. Took a bit to iron


jph00:
Sorry I don't understand this one. Can you please add some examples and details in the notebook and explain more details in the PR comment?

muellerzr:
Sure! Definitely rushed that, apologies.

Currently we have a large number of use with the `delegates()`, which let's us fill in **kwargs when we have the function signature and hinting. But we currently *don't* have this for `docments` and `show_doc`. It is just ignored, even when say docments might be helpful for knowing what can be passed in and how! 

This PR adds the start of this capability (which I'm hoping we can discuss logistics for this here)

For start, the benefit immediately:

This let's `show_doc` check what we `delegate`'d, see if we have any docments on whatever func was delegated, and fill in those docments automatically, under a new section in `show_doc`, that **Valid Keyword Arguments** section. *Otherwise* we state where that kwarg gets passed to. 

For display purposes, we have 3 scenarios, shown with the following:

## Scenario 1
Scenario 1 is where we just have `**kwargs` and a `docment`, and we want to fill in a table containing what `a` should be
```python
from fastcore.meta import delegates
def _a(
    a:int=2, # First
):
    return a

@delegates(_a)
def _b(
    b:str, # Second
    **kwargs
):
    return b, (_a(**kwargs))
```

This is straightforward, with `show_doc` providing: 
image

But what if `a` didn't have a docment? Then we tell the user where it gets passed to
image

(@jph00 I'll let you decide if we want to get rid of this. If we do that's an easy fix)

## Scenario 2
In scenario 2, var `b` has a docment, and I don't want to retype `b`'s docments. If I change what `b` does in `_c`, that's documentation I need to refactor. Instead we can automatically pull `b`'s docment from `_c`
```python
def _c(
    b:str, # Second
    a:int=2, # Blah
):
    return b, a

@delegates(_c)
def _d(
    c:int, # First
    b:str,
    **kwargs
):
    return c, _c(b, **kwargs)
```
image

Now the only "bug" with this one, is since `b` didn't originally start with a docment, it's thrown as a kwarg, showing that this is where that gets sent to. That can be fixed, but for this demonstration let's just go with this assumption

## Scenario 3
Picking up off scenario 2, if I write the following:
```python
@delegates(_c)
def _e(
    c:int, # First
    b:str, # Second
    **kwargs
):
    return c, _c(b, **kwargs)
```
We want to override the docment from `_c` for `b`, since `b` is a required delegated param, meaning we should have to document what it does (given the rules of `delegates`). As a result we render like so:
image


Does this make it clearer @jph00?

muellerzr:
The examples I just pulled are these tests:
```python
_str = '<h4 id="_b" class="doc_header"><code>_b</code><a href="__main__.py#L8" class="source_link" style="float:right">[source]</a></h4>\n\n> <code>_b</code>(**`b`**:`str`, **`a`**:`int`=*`2`*)\n\n\n\n||Type|Default|Details|\n|---|---|---|---|\n|**`b`**|`str`||Second|\n|||**Valid Keyword Arguments**||\n|**`a`**|`int`|`2`|First passed to `_a`|\n'
test_eq(show_doc(_b, disp=False), _str)
_str = '<h4 id="_d" class="doc_header"><code>_d</code><a href="__main__.py#L20" class="source_link" style="float:right">[source]</a></h4>\n\n> <code>_d</code>(**`c`**:`int`, **`b`**:`str`, **`a`**:`int`=*`2`*)\n\n\n\n||Type|Default|Details|\n|---|---|---|---|\n|**`c`**|`int`||First|\n|||**Valid Keyword Arguments**||\n|**`b`**|`str`||Second passed to `_c`|\n|**`a`**|`int`|`2`|Blah passed to `_c`|\n'
test_eq(show_doc(_d, disp=False), _str)
_str = '<h4 id="_e" class="doc_header"><code>_e</code><a href="__main__.py#L28" class="source_link" style="float:right">[source]</a></h4>\n\n> <code>_e</code>(**`c`**:`int`, **`b`**:`str`, **`a`**:`int`=*`2`*)\n\n\n\n||Type|Default|Details|\n|---|---|---|---|\n|**`c`**|`int`||First|\n|**`b`**|`str`||Second|\n|||**Valid Keyword Arguments**||\n|**`a`**|`int`|`2`|Blah passed to `_c`|\n'
test_eq(show_doc(_e, disp=False), _str)
```
Shown at the bottom of the `show_doc` notebook

muellerzr:
This is definitely a more complex change, so I appreciate all questioning you can give and I'll try and explain my decisions

jph00:
LGTM!

PR #593, opened 2022-02-26
Misc bug fixes + tests

muellerzr:
Fixes a variety of bugs pointed out in the discord including:
- Defaults of `None` showing as double backticks
- Parameter names link to types
Refactored slightly where we grab monospace.

(There may be a merge conflict with https://github.com/fastai/nbdev/pull/582 which should be merged before this. If so will address and fix)


muellerzr:
cc @jph00

jph00:
Cool @muellerzr - go ahead and fix up that conflict now

PR #594, opened 2022-02-28
Use the kernel defined in the kernelspec

dleen:
The notebook metadata has a kernelspec section which defines the kernel
to use with the notebook:
```
  ...
  "metadata": {
    "jupytext": {
     "split_at_heading": true
    },
    "kernelspec": {
     "display_name": "Python 3 (ipykernel)",
     "language": "python",
     "name": "python3"
    },
    ...
```

Instead of hard coding the kernel name to "python3" we use the default
behavior of using the kernel that the notebook was run with.

See: https://forums.fast.ai/t/modulenotfounderror-in-fresh-conda-environment/93666


jph00:
Many thanks!

jph00:
BTW for future reference, you need to run `nbdev_build_lib` after changing the nb. I've done that now.

PR #596, opened 2022-03-02
Better checks for cls and self

muellerzr:
Fixes a bug where `cls` and `self` were always popped if present, a lazy approach to testing if it's a method in a class  or not. This PR addresses this and provides a fix. Also fixes a slight error with `get_config` right now (see note in Discord on this). Should be merged after https://github.com/fastai/fastcore/pull/394


muellerzr:
CC @jph00 because oops forgot to!

PR #611, opened 2022-04-14
Change doc() default for docments

muellerzr:
# Change default for `doc` to not show docments always

## What does this add?

Slightly changes default behavior in `doc`

## Why is it needed?

Current default is to always show docments, even if no docments are present when doing `doc(elt)`, it should be `False` to mimic `show_doc` behavior and only show docments if they are available. Also removed :int typehint (unsure when this was added, dug through commits and still couldn't find where/when)

cc @jph00


jph00:
many thanks for fixing that!

PR #615, opened 2022-04-26
Added webrick spec to Gemfile.

MarkB2:
Building docker-compose image on WSL2 throws exception
`
nbdev-jekyll-1    | /var/lib/gems/3.0.0/gems/jekyll-3.9.2/lib/jekyll/commands/serve/servlet.rb:3:in require: cannot load such file -- webrick (LoadError)
`
Addition of
`
gem "webrick", "~> 1.7"
`
into `Gemfile` solves the problem.


muellerzr:
cc @hamelsmu probably? 😄

PR #621, opened 2022-05-05
Fix show doc for object, class methods.

v-ahuja:
The `skip_params` logic wasn't quite right. We're only ever supposed to skip the 1st param.


v-ahuja:
@jph00 the previous fix wasn't quite right. thanks.

jph00:
Makes sense - thanks

Issue #595, opened 2022-03-01
Including @dataclass breaks `nbdev_build_lib`


jvivian:
Version: 1.2.2

**Summary of issue**
Including a `@dataclass` decorated class breaks `getsource` in `show_doc`

**Minimal example**
- Clone `nbdev_tutorial`
- Add new cell in `00_core.ipynb`

```python
#export
from dataclasses import dataclass

@dataclass
class Foo:
    a: int
    b: str
    c: float
```
- Run `nbdev_build_lib`
- Run `nbdev_build_docs`

```
nbdev_build_docs
converting: /Users/jvivian/Dropbox/Programming/nbdev_tutorial/00_core.ipynb
An error occurred while executing the following cell:
------------------
show_doc(Foo, default_cls_level=2, show_all_docments=False)
------------------

---------------------------------------------------------------------------
OSError                                   Traceback (most recent call last)
/var/folders/6h/q9jpcd994cxbfl2rfsfc5wqwt2x66s/T/ipykernel_11200/3026477274.py in <module>
----> 1 show_doc(Foo, default_cls_level=2, show_all_docments=False)

~/anaconda3/envs/nbdev/lib/python3.7/site-packages/nbdev/showdoc.py in show_doc(elt, doc_string, name, title_level, disp, default_cls_level, show_all_docments, verbose)
    347             elt = elt.__init__
    348         if is_source_available(elt):
--> 349             if show_all_docments or _has_docment(elt):
    350                 s = f"\n\n{_format_args(elt)}"
    351                 if not monospace: s = add_doc_links(s)

~/anaconda3/envs/nbdev/lib/python3.7/site-packages/nbdev/showdoc.py in _has_docment(elt)
    238 # Cell
    239 def _has_docment(elt):
--> 240     comments = {o.start[0]:_clean_comment(o.string) for o in _tokens(elt) if o.type==COMMENT}
    241     params = _param_locs(elt, returns=True)
    242     comments = [_get_comment(line,arg,comments,params) for line,arg in params.items()]

~/anaconda3/envs/nbdev/lib/python3.7/site-packages/fastcore/docments.py in _tokens(s)
     21 def _tokens(s):
     22     "Tokenize Python code in string or function object `s`"
---> 23     if isfunction(s): s = getsource(s)
     24     return tokenize(BytesIO(s.encode('utf-8')).readline)
     25

~/anaconda3/envs/nbdev/lib/python3.7/inspect.py in getsource(object)
[...]

PR #622, opened 2022-05-08
Fix show_doc for dataclass

MarkB2:
`show_doc` couldn't show dataclasses (see for example #595). The problem is to show class, nbdev actually shows its constructor and dataclass doesn't have an accessible one. The PR temporary fixes this. The remaining problem is class level docments and class constructor arguments docments are not shown. This needs to be treated in fastcore.docments.


MarkB2:
Updated show_doc to reflect `fastcore.docments` update. cc @jph00

MarkB2:
Thanks.

PR #629, opened 2022-06-19
Fix logic for skipping params during function doc gen

v-ahuja:
For a regular function that's not part of a class:

```python
def some_func(cls, a, b):
      # logic
```

the generated signature in the documentation should include `cls` (or `self`). These keywords should only be skipped when a function is part of a class.


v-ahuja:
@jph00 tks

PR #635, opened 2022-06-27
fix show_doc for properties

seeM:
This fixes `show_doc` for properties, and adds a smoke test. There's a missing `else` branch in the nested if statements.

nbdev + fastcore + fastai tests pass locally.

cc @hamelsmu @jph00


jph00:
and clearer to read now - thanks! :D

PR #636, opened 2022-06-27
support py310 style union annotations

seeM:
This extends `nbdev.export` to work with `patch`ed functions that use py310 style union annotations, and converts all instances of the tuple style to py310 style. Tuple style still is still supported.

cc @hamelsmu @jph00


seeM:
@jph00 this is now ready for review. It extends `nbdev.export` to work with patched functions that use py310 style union annotations, and converts all instances of the tuple style to py310 style. The old style still works though, so this isn't dependent on any other changes to fastcore.

PR #657, opened 2022-07-24
doc consistency

seeM:
Apologies if this is nitpicky! I think it's important since `nbdev_help` is front-facing in the readme

cc @hamelsmu @jph00


hamelsmu:
Thanks!  Looks good

PR #670, opened 2022-07-27
Add early return

dleen:
It seems like there is a missing return statement when reading from stdin.

The current code reads from stdin and writes to stdout, but then proceeds to
running through all notebook files via globbing and processing all of them.

When git builds the index (happens even during git status) it passes all
files to the filter each via stdin which leads to n^2 behavior where n
is the number of notebooks.

@jph00


jph00:
Well spotted!

PR #682, opened 2022-07-29
Fix nbdev_install_hooks references in the docs

KBodolai:
Changed a couple of references to the old `nbdev_install_git_hooks` instead of `nbdev_install_hooks`.

Some issue still remains when previewing the documentation with `nbdev_preview`, all notebooks are stripped of the numbering prefix, except for `11_clean.ipynb`. (See screenshots below), which breaks the autogenerated cross references in the docs (which point to `clean.html` instead of `11_clean.html`)
Screenshot 2022-07-29 at 12 25 55
Screenshot 2022-07-29 at 12 27 14

I haven't been able to locate why that's going on yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,7 +133,8 @@ strive to abide by generally accepted best practices in open-source
 software development.
 
 Make sure you have `nbdev`’s git hooks installed by running
-`nbdev_install_git_hooks` in the cloned repository.
+[nbdev_install_hooks](https://nbdev.fast.ai/clean.html#nbdev_install_hooks)
+in the cloned repository.
 
 ## Copyright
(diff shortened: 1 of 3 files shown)


jph00:
Many thanks! I'll look into that odd `11_clean` issue.

KBodolai:
It seems to be working now, and the link points in the right direction in the docs. What was the issue?

jph00:
That nb is the only one with a raw cell containing frontmatter, instead of using an H1 in the markdown. There was no filename in the frontmatter.

KBodolai:
interesting, thanks!

PR #694, opened 2022-07-30
Conditional content for markdown vs HTML for README

hamelsmu:
The embedded youtube doesn't work in GitHub markdown, so we can use a nifty Quarto feature for [conditional content](https://quarto.org/docs/authoring/conditional.html) depending on whether something is HTML or Markdown!  This is because you cannot embed youtube in GitHub markdown

Also, we can use the full power of pandoc markdown instead of just regular markdown and Quarto will figure out how to convert it!  In this situation, I have made the video open in a a new tab with `{target="_blank"}`

```
[![](https://github.com/fastai/logos/raw/main/nbdev_walkthrough.png)](http://www.youtube.com/watch?v=l7zS8Ld4_iA "nbdev walkthrough"){target="_blank"}
```

I also fixed some missing content in the notebook

cc: @jph00 @seeM


jph00:
Very cool @hamelsmu !

PR #707, opened 2022-08-01
deterministic `show_doc` and `DocmentTbl` `repr`

seeM:
@hamelsmu @jph00 what do you think about this for less polluted diffs?


jph00:
Excellent idea!

PR #717, opened 2022-08-02
Add nbs_path migration note

DanteOz:
Added note to migration guide indicating that styles.css needs to be moved inside of nbs_path if there is a custom nbs_path in settings.ini.

### DIFF
--- a/nbs/migrating.ipynb
+++ b/nbs/migrating.ipynb
@@ -123,7 +123,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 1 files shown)


nbs/migrating.ipynb
```diff
@@ -114,7 +123,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "A number of directives have changed names. We'll use `perl` to fix them. Run these lines in the root of your repo:\n",
+    "If you have set an `nbs_path` in your `settings.ini` file, then copy your styles.css file inside of your `nbs_path` folder.A number of directives have changed names. We'll use `perl` to fix them. Run these lines in the root of your repo:\n",
```
hamelsmu, line 126:
```suggestion
    "A number of directives have changed names. We'll use `perl` to fix them. Run these lines in the root of your repo:\n",
```

PR #732, opened 2022-08-03
non-notebooks do not have nbformat field

dleen:
The tuple comparison fails on files which aren't notebooks. First we
need to check the type, then we need to check the version


dleen:
@jph00

jph00:
Many thanks!

PR #738, opened 2022-08-03
add `user` option to `jupyter_hooks` setting

seeM:
Note that this isn't yet useful. I'm considering changing `get_config` to read `~/.nbdev/settings.ini` if it exists, then `update` it with `$(pwd)/settings.ini`, then `jupyter_hooks=user` will be useful. @hamelsmu @jph00 what do you think?

### DIFF
--- a/nbdev/clean.py
+++ b/nbdev/clean.py
@@ -122,14 +122,21 @@ def nbdev_clean(
 # %% ../nbs/11_clean.ipynb 24
 def clean_jupyter(path, model, **kwargs):
     "Clean Jupyter `model` pre save to `path`"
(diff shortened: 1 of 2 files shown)


jph00:
@seeM how about for the location we follow the [XDG spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html)?

> $XDG_CONFIG_HOME defines the base directory relative to which user-specific configuration files should be stored. If $XDG_CONFIG_HOME is either not set or empty, a default equal to $HOME/.config should be used.

jph00:
I've just added a new module `fastcore.xdg` to make this more convenient.

PR #756, opened 2022-08-04
fix #747

### LINKED ISSUE #747: Streamline default settings
- We shouldn't have to specify each config key's default every time we use it
- Config keys (and their defaults) should all be documented in one place

seeM:


seeM:
@hamelsmu @jph00 I'd like to hear your thoughts on this PR. IMO, upsides:

- no need to put defaults in every `config_key`, which should also make it easier to keep tools working outside of nbdev repos moving forward
- all defaults are defined and documented in one place

Downsides:

- the `show_doc` of `nbdev_create_config` (which this PR suggests should serve as the main reference for nbdev settings) is polluted by 'passed to `apply_defaults`' - we could add an option to `delegates` to remove that

seeM:
Thought of another downside to calling `apply_defaults` in `get_config`: we wouldn't be able to use a different default value in `config_key` if we wanted to. I don't think `get_config` should call `apply_defaults`, we can rather call it when we need it in specific commands.

I also feel like there must be a simpler way to do all this but am a bit stumped..

jph00:
Looks like a good step to me! :)

PR #767, opened 2022-08-05
add site url

hamelsmu:
For preview images and social cards: https://quarto.org/docs/websites/website-tools.html#preview-images


hamelsmu:
For preview images and social cards: https://quarto.org/docs/websites/website-tools.html#preview-images

PR #770, opened 2022-08-06
use global defaults instead of respecifying each time

seeM:


seeM:
@hamelsmu @jph00 while we're at it, I probably should've tagged you both before merging this one too, please have a look

PR #776, opened 2022-08-07
preview quarto default to random port

tourdownunder:
To avoid

```sh
nbdev_preview 
```

```sh
ERROR: Requested port 3000 is already in use.
```

Its seems neat to just omitted the port and [preview](https://quarto.org/docs/reference/projects/options.html#preview) with by default Port to random value between 3000 and 8000).

### DIFF
--- a/nbdev/cli.py
+++ b/nbdev/cli.py
@@ -221,7 +221,6 @@ project:
   type: website
   output-dir: {doc_path}
   preview:
-    port: 3000
     browser: false
 
 format:


jph00:
No we prefer the static port. But feel free to override it in custom.yml

tourdownunder:
Thanks creating `custom.yml` in my project that uses `nbdev` v2 is all I needed to do. With:

```yml
project:
  preview:
    port: 3003
```

tourdownunder:
Thanks creating `custom.yml` in my project that uses `nbdev` v2 is all I needed to do. With:

```yml
project:
  preview:
    port: 3003
```

PR #778, opened 2022-08-07
Allow users to provide user-level settings in ~/.config/nbdev/settings.ini

### LINKED ISSUE #755: User-level settings
via @jph00 

> how about for the location we follow the [XDG spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html)?
>
> > $XDG_CONFIG_HOME defines the base directory relative to which user-specific configuration files should be stored. If $XDG_CONFIG_HOME is either not set or empty, a default equal to $HOME/.config should be used.
>
> I've just added a new module fastcore.xdg to make this more convenient.

seeM:
This allows users to provide user-level settings in `~/.config/nbdev/settings.ini` by default, but fully supporting the XDG spec thanks to [`fastcore.xdg`](https://fastcore.fast.ai/xdg.html). The main use-case is to configure nbdev tools outside nbdev repos. Settings are read in order of: defaults, user, repo (later overriding earlier).

cc @hamelsmu @jph00


seeM:
@jph00 it looks like this needs a fastcore release

PR #784, opened 2022-08-07
Hooks ignore settings.ini if Jupyter is started outside nbdev repo

dleen:
Jupyter hooks are run from Jupyter's cwd (can be specified or else directory Jupyter was launched from). Test this by launching Jupyter from outside your current project's directory.

When this isn't the same directory as the nbdev repos root directory the settings file is not found.
Because some settings have default values e.g. allowed_metadata_keys the hook doesn't fail when the settings file is missing.

Instead we look at the path of the notebook given to the hook and attempt to find the settings file by
traversing directories upward until we find it or hit root '/'.

@jph00 @seeM

### DIFF
--- a/nbdev/clean.py
+++ b/nbdev/clean.py
@@ -98,9 +98,10 @@ def process_write(warn_msg, proc_nb, f_in, f_out=None, disp=False):
         warn(e)
 
 # %% ../nbs/11_clean.ipynb 23
-def _nbdev_clean(nb, **kwargs):
-    allowed_metadata_keys = config_key("allowed_metadata_keys", path=False).split()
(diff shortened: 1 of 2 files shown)


seeM:
Ah, thanks @dleen! IIUC this was a regression on my part today. Previously we were using `get_config` instead of `config_key` which already searches up, and which can be passed a path to search from. If I’m correct, would you mind updating the PR to use `get_config`?

seeM:
BTW I’m going to be working with this code in the next hour, so very happy to take it from here

dleen:
Much easier knowing that `get_config` searches up! @seeM

seeM:
Perfect 😄 thank you @dleen!

PR #788, opened 2022-08-07
Clean `id` from text `repr` outputs to further avoid git merge conflicts

seeM:
It can be disabled using `clean_ids = False` in user- or project-settings. It can't yet be controlled per-notebook. I'd appreciate some guidance on how best to implement that, and thought it could be done in a follow-up PR.

Also tested by running `nbdev_clean` on fastai, then checking that `rg ' at 0x[0-9a-fA-F]+' nbs` finds nothing.

cc @hamelsmu @jph00


jph00:
Terrific!

PR #795, opened 2022-08-08
links aren't rendered as code

seeM:
cc @hamelsmu @jph00


jph00:
I think you might have forgotten to export

hamelsmu:
Hmm CI should be catching the non synced notebook I'll look

PR #797, opened 2022-08-08
fix typo in migrating nb

deven367:
In the migrating notebook, the instruction had a typo, `nbdev_migrate_directives` was changed to `nbdev_migrate`


ulupo:
I am using 2.0.7 from the fastai conda channel, and the command is still `nbdev_migrate_directives`.  Perhaps consider mentioning both commands in the documentation?

PR #812, opened 2022-08-09
refactor tutorial intro and installation

seeM:
Happy to discuss whether you think this is an improvement or not, particularly cutting the intro so much. I think the exploratory programming stuff is better shown than told. I figure people reading this want to see, practically, how to use nbdev. I've tried to cater for absolute beginners. cc @hamelsmu @jph00


hamelsmu:
LGTM!

PR #813, opened 2022-08-09
nbdev commands fail when `doc_path` contains whitespace

mone27:
If `doc_path` contains a white space, `nbdev_quarto` crashed as there were no quotes around in the `cd` command.
The same for `nbs_path` and `_render_readme`.

According to the contributing guide every PR should include a test. However, the change in minimal and properly adding a test would require some effort to properly test that the library works with white spaces so I was wondering if it is really needed in this case.

### DIFF
--- a/nbdev/cli.py
+++ b/nbdev/cli.py
@@ -320,7 +320,7 @@ def _render_readme(path):
         yml_path.rename(path/'sidebar.yml.bak')
         moved=True
     try:
-        _sprun(f'cd {path} && quarto render {idx_path} -o README.md -t gfm --no-execute')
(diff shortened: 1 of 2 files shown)


seeM:
Thanks, this is a helpful contribution! I agree that a test isn't needed here

It looks like CI is failing (https://github.com/fastai/nbdev/runs/7750171492?check_suite_focus=true) with:

```
Error: Notebooks and library are not in sync.
```

Could you please run `nbdev_export` and `cli.py` push the changes?

mone27:
ok!
Sorry for not running `nbdev_export`, now it should done

Issue #472, opened 2021-04-26
A command for uploading to the test PyPi server?


tirthajyoti:
Can you guys add a command for uploading to the [Test PyPi server](https://test.pypi.org/) instead of the main one? Surely, this will reduce the clutter on the main server that many may complain about.

Of course, a separate set of credentials must be added to the config file by the user for this to work.

tourdownunder:
Hi @tirthajyoti 

I made a few small changes to get this in tonight in https://github.com/tourdownunder/nbdev/pull/1. I was able to successfully with my gitlab repo that has a private package index.

What do you think?

tirthajyoti:
Thx!

PR #818, opened 2022-08-09
#472 a command for uploading to the test pypi server

tourdownunder:
To solve #472 by passing `--repository` option to twine to upload pypi, testpypi or other location defined in ~/.pypirc

Tested manually using the following following with success:

```sh
nbdev_pypi --repository gitlab-python-project
```

Where `~/.pypirc` has entry:

```sh
[gitlab-python-project]
repository = https://gitlab.com/api/v4/projects/XXXXXX/packages/pypi
username = __token__
password = glpat-XXXXXXXXXXXXX
```

Unsure how to unit test this.


jph00:
Thanks @tourdownunder . I think you might have forgotten to run `nbdev_prepare` before you pushed, since I don't see any changes to the .py module. Would you be able to do that please? (That's also why you see the CI failures here)

nbs/16_shortcuts.ipynb
```diff
@@ -258,44 +262,7 @@
    "execution_count": null,
    "id": "b50d27fb-c343-47e2-a008-7ebdd02e1356",
    "metadata": {},
-   "outputs": [
```
jph00, line 261: For future reference, be careful about clearing outputs and then not rerunning cells, because this means it won't appear in the docs any more. Don't worry about it this time -- I'll do it -- this is just an FYI

PR #820, opened 2022-08-10
Streamline `nbdev_new`: outputs are now in color, you can pass `--lib_name`, and it calls `nbdev_export`

seeM:
Run `nbdev_export` removes one step from the minimal hello world tutorial. Here's what the colors looks like:

image

cc @hamelsmu @jph00

### DIFF
--- a/nbdev/cli.py
+++ b/nbdev/cli.py
@@ -170,6 +170,9 @@ def extract_tgz(url, dest='.'):
     with urlopen(url) as u: tarfile.open(mode='r:gz', fileobj=u).extractall(dest)
 
 # %% ../nbs/10_cli.ipynb 17
+def _mk_cfg(**kwargs): return {k: kwargs.get(k,None) for k in 'lib_name user branch author author_email keywords description repo'.split()}
+
+# %% ../nbs/10_cli.ipynb 18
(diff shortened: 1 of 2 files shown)


nbdev/cli.py
```diff
@@ -186,35 +189,41 @@ def _get_info(owner, repo, default_branch='main', default_kw='nbdev'):
...
+        res['author'] = run('git config --get user.name').strip()
+        res['author_email'] = run('git config --get user.email').strip()
+        res['user'],res['repo'] = repo_details(url)
+        res['branch'],res['keywords'],res['description'] = _get_info(owner=res['user'], repo=res['repo'])
     except OSError as e:
         if raise_err: raise(e)
-        return dict(lib_name=None,user=None,branch=None,author=None,author_email=None,keywords=None,description=None,repo=None)
-    return dict(lib_name=repo.replace('-', '_'), user=owner, branch=branch, author=author, 
-                author_email=email, keywords=keywords, description=descrip, repo=repo)
+    res['lib_name'] = res['repo'].replace('-','_')
```
seeM, line 204: Oops, this is a bug

jph00:
Good stuff! Feel free to merge when ready. We should probably add some minimal terminal formatting stuff to fastcore to make those ansi codes more readable...

PR #829, opened 2022-08-11
refactor tutorial: first steps

seeM:
I restructured into sections:

- Installation
- First steps
- Make your first edit
- Advanced functionality

And reworked the "First steps" section here.

@hamelsmu @jph00, should I look into compressing the images or are these sizes fine? Are we concerned about the repo becoming huge?


jph00:
Wow this is great! Yes I do think it's worth making these images smaller - the are much higher resolution than they need to be. Perhaps a max width of 1200 or so?

seeM:
Done

Issue #696, opened 2022-07-30
previewed doclinks point to localhost


seeM:

jph00:
Can you provide some more detail on this @seeM ? I'm not following what it's for.

seeM:
Currently, docs rendered locally with `nbdev_preview` have doclinks to nbdev.fast.ai instead of localhost

hdocmsu:
It may not be directly related to this issue but I am having an issue with the auto-generated hyperlinks of classes. I followed Jeremy's tutorial for nbdev_cards. Everything works beautifully except the hyperlink. For some reason the auto-generated hyperlinks have an additional "nbdev_cards." added to the beginning of file.html.

Specifically, when using `Card` in the markdown, I would expect the below hyperlink
https://hdocmsu.github.io/nbdev_cards/card.html#card

however, the hyperlink I got below has the additional prefix "nbdev_cards." on the card.html
https://hdocmsu.github.io/nbdev_cards/nbdev_cards.card.html#card

Any suggestions are very appreciated. Thanks!

hdocmsu:
One more comment, in nbdev1, I recalled that there is a hyperlink that leads to the source code in github.com. I don't see the link-to-the-source-code in nbdev2. I am not sure if this is intentional or a bug. Thanks!

seeM:
This is addressed in https://github.com/fastai/nbdev/pull/832

@hdocmsu are you still having an issue with your first question?

The second issue you mentioned is being tracked here: https://github.com/fastai/nbdev/issues/692

hdocmsu:
Hi @seeM, yes, I still have a problem with my first question. For some reason, there is an extra "nbdev_cards." added to the file.html. 

I think the fast.ai/nbdev_cards (https://github.com/fastai/nbdev_cards) also has the same issue with the hyperlink. Here is the hyperlink from 'Card' in the README.md from the repository (https://github.com/fastai/nbdev_cards). 

https://fastai.github.io/nbdev-cards/nbdev_cards.card.html#card

PR #832, opened 2022-08-11
Make doclinks use relative URLs

seeM:
This makes doclinks use relative URLs, which works since we include `site-url` in `_quarto.yml`, fixing #696. It's actually only a one-line change in `doclinks.py` - the rest are tests and modidx. cc @hamelsmu @jph00


jph00:
Does this impact nbdev-stdlib and the other indexes at all?

Does this impact the ability to use this project's index on other projects?

seeM:
Ah, I overlooked that! I've made this a draft PR while I look into it. My gut feeling is that it's not actually a quick fix then

seeM:
Actually I'll close and open a new PR when appropriate

PR #833, opened 2022-08-11
Upgrade fastcore for xdg

benoit-cty:
Previously pinned version is missing `xdg`, so upgrading nbdev from 2.1.1 to 2.1.3 result in error `ModuleNotFoundError: No module named 'fastcore.xdg'`


seeM:
Thanks @benoit-cty! It looks like you'll need to run `nbdev_prepare` to ensure your notebooks and library are in sync (and cleaned and tests pass) for CI to pass: https://github.com/fastai/nbdev/runs/7798580611?check_suite_focus=true

benoit-cty:
Hello,
Thanks, I've done it.

seeM:
Thank you for the contribution! :)

PR #839, opened 2022-08-12
skip install if quarto is on the path

### LINKED ISSUE #840: check if quarto is on the path
Using `shutil.which` we could quickly check if quarto is already there, to avoid unnecessary install.

michaelaye:
just a quick check with `shutli.which`

fixes #840


jph00:
This command is also used for upgrading quarto, so checking it's in the path breaks it.

PR #842, opened 2022-08-12
Add custom_quarto_yml setting

benoit-cty:
Quarto config could be customized by custom.yml but it does not work for all settings.

So I propose to add a nbdev settings to prevent overwriting _quarto.yml.

This option as to be `custom_quarto_yml = False` to keep _quarto.yml.


jph00:
> Quarto config could be customized by custom.yml but it does not work for all settings.

What settings does it not work for? Can you show us exactly what you tried, and what happened when you did it?

benoit-cty:
I tried to override the icon of my repo because I use Gitlab and not Github.
So I add in custom.yml:
```
website:
  navbar:
    right:
      - icon: git
        href: "https://git.leximpact.dev/leximpact/leximpact-prepare-data/tree/master/"
```

But it result in two icon in navbar.

jph00:
Makes perfect sense - thanks for explaining.

PR #847, opened 2022-08-13
Escape Footnotes from Docments Table

### LINKED ISSUE #796: `show_doc` paremeter default may render as footnote
For example:

https://github.com/fastai/nbdev/blob/cba39a3e178ec684c59c4bc1cc8a044e0e3af31d/nbs/04b_doclinks.ipynb?short_path=f0ae976#L415-L416

renders:

image

with footnote:

image

hamelsmu:
Closes #796 

@seeM @jph00

### DIFF
--- a/nbdev/showdoc.py
+++ b/nbdev/showdoc.py
@@ -29,14 +29,17 @@ def _bold(s): return f'**{s}**' if s.strip() else s
 def _escape_pipe(s): return re.sub(r'(\\)?\|', '\|', s)
 
 # %% ../nbs/08_showdoc.ipynb 9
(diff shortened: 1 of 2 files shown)


hamelsmu:
I tested it and this is how it renders

image

seeM:
LGTM, merge when ready!

Sharing in case it helps: I would've opted for a single function (maybe `_escape_md`) that ensures that a given list of characters are escaped - by doing something like `_escape_pipe` for each char. So far, that list would be `['|','^']`. It feels like we may hit this issue again, and that would simplify its solution

PR #848, opened 2022-08-13
Ignore .ipynb_checkpoints folder in module dir

dleen:
@jph00


jph00:
Thanks - that's the exact right solution.

PR #849, opened 2022-08-13
prettify output for `nbdev_test`

deven367:
If tests fail in `nbdev_test`, the console output would look something like this,

image

After the minor fix, it looks like this

image


jph00:
Thanks!

PR #852, opened 2022-08-13
Change order of git commands during setup: local first, global second & remove colors

### LINKED ISSUE #846: nbdev_new doesn't infer anything if no gitconfig
nbdev_new was assuming the existence of ~/.gitconfig but not saying anything when it didn't find it.

> Solution turned out to be that on my brand-new computer, there was no ~/.gitconfig file yet.  nbdev_new was relying on that but wasn't finding anything, and wasn't putting out any error message.

cc @drscotthawley

hamelsmu:
This fixes #846 , by changing the order in which the git commands run, such that the local repository is inspected first and the global user's gitconfig is inspected last.  That way, if this fails, we still have _some_ information we can infer on behalf of the user.  

Furthermore, this removes colors as it made the text invisible for me.

cc: @jph00 @seeM


seeM:
LGTM!

PR #856, opened 2022-08-14
deprecate `config_key` in favor of `get_config`

seeM:
cc @hamelsmu @jph00


seeM:
Will fix later today

jph00:
Great idea to add the deprecation warning - thanks!

PR #858, opened 2022-08-14
revert remove `nbs/17_release.ipynb`

seeM:
https://github.com/fastai/nbdev/commit/d32a791a3a6929e9d22fe91e0f48cd330c39a339 accidentally removed `nbs/17_release.ipynb`. This adds it back. I'm not sure if there's a way to get the old git history, maybe [`--ignore-revs-file`](https://git-scm.com/docs/git-blame#Documentation/git-blame.txt---ignore-revs-fileltfilegt). cc @hamelsmu @jph00

### DIFF
--- a/nbs/17_release.ipynb
+++ b/nbs/17_release.ipynb
@@ -1,0 +1,769 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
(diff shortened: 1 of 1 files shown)


jph00:
Thank you!

PR #862, opened 2022-08-15
`show_doc` treats functions decorated with `lru_cache` as classes

seeM:
@hamelsmu @jph00, not sure if this is a good solution?


seeM:
Maybe a better solution is to use `sym.__wrapped__` if it exists

seeM:
Just realised `show_doc` also isn't showing a param table for these functions

jph00:
In general I suspect we should be using `__wrapped__` when it's available.

jph00:
I've just changed it to use `__wrapped__` instead - seems to work OK.

PR #864, opened 2022-08-15
Improve config documentation in `read` module

seeM:
This adds documentation to config-related code in the `read` module. It shouldn't have any code changes (except a one-liner fix in `nbdev_create_config`. cc @hamelsmu @jph00


hamelsmu:
LGTM

PR #878, opened 2022-08-16
improvements to `nbdev_new` and `nbdev_create_config`

seeM:
- `nbdev_create_config` infers settings from git/GitHub, prompts for missing settings, and renders the settings file with commented sections
- `nbdev_new` uses `nbdev_create_config` instead of a file provided by `nbdev-template`, which means it'll benefit from future improvements to `nbdev_create_config` as well as always using latest defaults


seeM:
@hamelsmu @jph00 this is ready for review. Note that most of the additions in `config` are cut-paste from `cli`

PR #910, opened 2022-08-20
Fix missing rename read->config

dleen:
@jph00


jph00:
Thanks @dleen  -- I just noticed that myself and had fixed it at my end!

Issue #885, opened 2022-08-17
"Go to definition" with `@patch`


dleen:
Sometimes I use VSCode when editing notebooks. The "intellisense" features (powered by pylance/pyright) work surprisingly well.

I really like using `@patch` when adding methods to functions in notebooks as it really helps with the flow of the notebook and the ability to write prose.

Because `@patch` dynamically adds methods, features like "Go to definition" do not work. I believe that `__dir__` does correctly present the right methods when writing a dot e.g. `f = Foo(); f.<completion>`.

I wanted to open this issue to see if there's any interest in getting something like this to work. I made the issue in nbdev instead of fastcore as I think there's a chance that some sort of nbdev processor could run through the notebooks, detect the patch annotation, and possibly generate stub files to let the IDE or completion tool e.g. Jedi known what methods have been added.

Just want to open this up to discussion. Thoughts?

jph00:
I think it would be great if that worked! I'm not sure it's even possible though. If you can create some POC, that would be very interesting to look at.

jph00:
@dleen once #1000 and #1001 are done, this should show clearly how to do this as an external plugin module. Does that sound like a reasonable next step for this?

dleen:
Yep - sounds perfect! Are you planning on tackling #1000 ?

jph00:
FYI @dleen this is actually implemented now - `py2pyi` generates a pyi file.

PR #928, opened 2022-08-24
POC: `@patch` annotation processor for early feedback

dleen:
See: #885. @jph00 sending this out for early feedback. Not sure if this is something that `nbdev` actually wants, but would like to hear your thoughts.

The central code is found in `03_process.ipynb` in `PatchAnnotationProcessor`, and all the rest of the changes in the `nbdev` files are the result of running the processor across nbdev itself to get a feel for what the processor does. I confirmed that code completion/type checking tools indeed work (when you import a module, not within the module you are editing, but that doesn't matter as much since you have the `@patch` methods right in front of you)


nbs/02_maker.ipynb
```diff
@@ -248,6 +248,7 @@
     "#|export\n",
```
jph00, line 248: <p>When is this needed?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/fastai/nbdev/pull/928/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='18'/>

nbs/03_process.ipynb
```diff
@@ -32,11 +32,16 @@
     "from nbdev.maker import *\n",
```
jph00, line 32: <p>I'd rather avoid dataclasses if possible.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/fastai/nbdev/pull/928/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='44'/>

jph00, line 32: <p>Markdown is best used for user-facing docs of what a class does, rather than how it works. Comments are generally best for documenting how it works - ideally at the site it occurs.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/fastai/nbdev/pull/928/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='45'/>

nbs/04a_export.ipynb
```diff
@@ -133,7 +133,7 @@
    "outputs": [],
```
jph00, line 133: <p>Carefuly of python mutable parameters!</p><p>https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/fastai/nbdev/pull/928/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

nbs/04b_doclinks.ipynb
```diff
@@ -33,7 +33,7 @@
     "from fastcore.utils import *\n",
```
jph00, line 33: <p>Is this needed for this particular PR? AFAICT it's just to avoid some possible problem where the marker is inside a multi-line string?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/fastai/nbdev/pull/928/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

tests/patch.ipynb
```diff
@@ -0,0 +1,118 @@
+{
```
jph00, line 1: <p>I'd rather avoid having a whole new nb just to test one feature if possible</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/fastai/nbdev/pull/928/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

jph00:
This is very interesting! I don't know whether we should do this or not. If we do, I suspect that some (or most?) of the logic should move into fastcore.

@seeM @hamelsmu I'd love to hear your thoughts.

jph00:
Another idea would be to make this into an external project, and use it to ensure that it's easy for users to add their own processors to their projects.

jph00:
I just realised that it might be impossible to get 2-way sync working with this processor...

nbs/04b_doclinks.ipynb
dleen: <p>It's needed because rewriting the source code from the ast in this method: https://github.com/fastai/nbdev/blob/master/nbdev/maker.py#L209 causes some reformatting from a single line to a multiline and then it's treated as a separate notebook cell. I was thinking of submitting this as a separate patch, but it's not needed unless someone changes the formatting of that file.</p>

jph00: I'd rather avoid adding such complex code unless it's definitely required! And if we really do need it, we should include plenty of comments explaining what it's doing and why. Or better still, find a way to avoid it.

nbs/02_maker.ipynb
dleen: <p>If you use <code>@fastcore.basics.patch</code> (or whatever the right import is) instead of importing directly using <code>from fastcore.basics import *</code> and then using <code>@patch</code> </p>

nbs/04b_doclinks.ipynb
dleen: <p>Yes, the error raised by this was a bit of head scratcher and I had to do a bit of digging on <code>ast</code> and newlines. As an example: https://stackoverflow.com/questions/53636723/python-parsing-code-with-new-line-character-in-them-using-ast.</p><p> </p><p>This is only an issue if:</p><ol><li>someone tries to rewrite that line as a triple quoted string using actual newlines</li><li>anything happens to parse that method using ast and rewrite it (which happens here since it has the @patch annotation). </li></ol><p> </p><p>I can guarantee you that I had exactly zero fun figuring out this regex, and I only did it to make progess on the actual problem that I wanted to solve (making @patch more usable). I'll leave it up to you to decide.</p>

dleen:
> I just realised that it might be impossible to get 2-way sync working with this processor...

I had thought about how to accomplish this but I didn't want to solve every problem otherwise this POC was going to take forever, and I really just wanted feedback about the viability of the approach, and even if you consider it a problem worth solving in the first place. It would be terrible to spend many hours on this for it to be thrown away. But this PR didn't take too long so I don't mind if it's not merged

To get the 2-way sync working again I could just use a comment annotation on the method (e.g. `# %% ../nbs/09_API/04a_export.ipynb 7 @patch`) telling nbdev where it originated from, similar to how the source code is matched up with the notebook cells.

dleen:
> Another idea would be to make this into an external project, and use it to ensure that it's easy for users to add their own processors to their projects.

Yes, I had this thought too and I might explore it a bit.

dleen:
> One thing I've found myself doing recently is using @patch to build my class interactively, but then merge the cells together when it's working into a single class. It can be nice to see everything together like that, at least for small-ish classes.

Oh this is interesting. Why have you started doing this recently?

jph00:
> To get the 2-way sync working again I could just use a comment annotation on the method (e.g. `# %% ../nbs/09_API/04a_export.ipynb 7 @patch`) telling nbdev where it originated from, similar to how the source code is matched up with the notebook cells.
> 

Ah yes OK - makes sense!

dleen:
@jph00 I appreciate you taking the time to review this, and for leaving valuable comments! I just want to reiterate that there's no pressure to merge this - I had a lot of fun writing a processor and learned a lot about the internals of nbdev in the process.

I am new to things like `@patch` and the fastcore "dialect" of Python and hopefully my fresh eyes can offer some insight from a beginners perspective. I found methods from fastcore defined using `@patch` quite difficult to track down even with the doc pages open (I would find myself clicking through each page doing a ctrl-f to find something), and my IDE was not much help either (I like to write code using neovim - but I usually have some sort of IDE open if I really want to dive into the internals of something and click through definition after definition).

jph00:
> Oh this is interesting. Why have you started doing this recently?
> 

Take a look, for instance, at how the `process` and `maker` modules used to be defined:

https://github.com/fastai/nbdev/blob/336598ac8953dd5c0ec18922fb06e907ef07ca52/nbs/02_export.ipynb

Everything was in separate cells, and I found it difficult to get my head around how they all worked together. Building it up in this way worked well for me, but maintaining it (especially after a break) didn't work so well. Once I put everything together so I could see it without scrolling, I found it easier to keep it all in my head and see how to improve it.

seeM:
This is very cool David! I’d definitely try it if/when I use a notebook editor with LSP support. I think it should be opt-in though, because I suspect it will have teething issues since it’s a significant change

Assuming that in future it works with everything (2-way sync, etc), would there be any downside to making this the default?

seeM:
I also like the idea of using this to dogfood nbdev extensibility. Though of course that would require more work. I think @hamelsmu was considering customising nbdev’s core behaviour at some point

dleen:
> Assuming that in future it works with everything (2-way sync, etc), would there be any downside to making this the default?

No, I think it makes `@patch` even more useful by removing its main disadvantage (the dynamic nature) and increases the utility of nbdev: you can safely improve productivity in notebooks by adding more and more to the Python "dialect" by leveraging dynamic features of Python (in this case runtime patching of classes) while "processing" notebooks into more "standard" Python modules which don't need to leverage as much of the "dialect" since it's no longer in a notebook.

seeM:
Hmm that’s a fascinating idea! It’s extending the dialect (started by fastcore) to language syntax and compilation. And since it compiles to more traditional-looking Python, it benefits from its ecosystem (especially with 2way sync)

Some loose thoughts off the top of my head:

One downside that comes to mind is that it’s more complex (I think) and might confuse beginners. I don’t think your patch extension is particularly difficult, provided it fails well with clear messages/diagnostics. I guess this would be up to us to ensure the new dialect’s “compiler” is delightful and beginner-friendly

I wonder if there’s a benefit to the dialect being aimed at notebooks in particular, rather than compiling from plaintext? I suspect there would be advantages afforded by notebooks, even if we aren’t leveraging them yet (are we?)

jph00:
Let's go ahead and make this an nbdev extension, if you're up for it. Then we'll all be able to try it out, and see how it goes, and potentially merge it back into the main project in the future. So I guess I should close this issue.

> I am new to things like @patch and the fastcore "dialect" of Python and hopefully my fresh eyes can offer some insight from a beginners perspective. I found methods from fastcore defined using @patch quite difficult to track down even with the doc pages open (I would find myself clicking through each page doing a ctrl-f to find something), and my IDE was not much help either

@dleen I find your insights *extremely* helpful. It might be worth jumping on a call to discuss this more f2f, but in the meantime I'll jot down some thoughts here. I should say though -- we're *all* pretty new to this way of working, and we're figuring out stuff as we go! I guess I've been doing this the longest, but I still feel like I've got a lot of things I'd like to improve. I'm many times more productive than I was when using TDD and vim/vs/pycharm/etc, but I suspect I'm only scratching the surface still.

In general, the fast.ai/nbdev smooth path is to do nearly everything in notebooks. One foundational premise is that python is a dynamic language, and that this can be leveraged to write more expressive, concise, and clear code more easily than in static language. The python stdlib takes advantage of this, for instance, to give us stuff like dataclasses. However, python's dynamic features are not compatible with static tooling, such as nearly all IDEs and editors. Instead, IDEs have to put in place hacks to recognise stuff like dataclasses.

OTOH Jupyter doesn't have to "guess" what a symbol is or where it comes from. When you're coding interactively, it *knows*, because the symbol is attached to a real object. So to find details about the type of an object, or the signature/docstring of a method, type `object?` or `object.method?`. To see the source, use `??` instead of `?`. Use `dir(obj)` to see what attrs it has. You can see exactly what its value is, and how it responds to different methods. And so forth...

My way of working nowadays looks *very* different to what I did in the pre-nbdev days! I generally put a 1-2 lines of code into a cell, evaluate it, check it's what I want, and then write another cell to further process that result, and so forth. Once I've got a bunch of cells that have done the thing I want, I merge them, add a function/class/whatever header, and then write some examples/tests.

I'm very interested in things that make it easier to use IDEs with fast.ai's libraries. But I'm also interested in finding ways to better explain how to fully take advantage of notebook-based development!

PR #930, opened 2022-08-24
Handle repo names with dashes and correct index page rendering with file attachments

### LINKED ISSUE #924: IPython.display.Image(embed=True) results in incorrect image reference in GitHub Pages
## To reproduce:

1. Add a code block that embeds an image from a remote URL.
```
from IPython.display import Image
Image(url="https://avatars.githubusercontent.com/u/20547620", format='png', embed=True)
```

2. Note that it renders correctly when using nbdev_preview:
Screen Shot 2022-08-23 at 11 50 24 AM

3. After pushing to GitHub, the image links are broken:
Screen Shot 2022-08-23 at 11 50 48 AM

The HTML page references a /figure-**html** folder:
`github.io/.../index_files/figure-html/cell-11-output-1.png`

while the actual location of the file is in a /figure-**gfm** folder:
`github.io/.../index_files/figure-gfm/cell-11-output-1.png`

hamelsmu:
Fixes #924 

Issues fixed:
- dashes in repo name will not break links
- Markdown attachments for README are moved to the root
- local index page attachments will not result in broken links.

cc: @seeM @jph00


nbdev/config.py
```diff
@@ -67,7 +67,8 @@ def _apply_defaults(
     if copyright is None: copyright = f"{datetime.now().year} ownwards, %(author)s"
     for k,v in locals().items():
         if k.startswith('_') or k == 'cfg' or cfg.get(k) is not None: continue
-        cfg[k] = v
+        if k == 'lib_name': cfg[k] = cfg.repo.replace('-', '_') # the repo name might have dashes but the lib name cannot
```
seeM, line 70: Shouldn’t lib_name default to ‘%(repo)s’ as it used to? Dashes are fine in a package name just not in the importable package itself, which is lib_path (already handled above)

nbdev/config.py
```diff
@@ -29,17 +29,17 @@ def _git_repo():
 # %% ../nbs/01_config.ipynb 10
 def _apply_defaults(
     cfg,
-    lib_name='%(repo)s', # Package name
+    lib_name:str=None, # Package name
```
seeM, line 32: I think this default was fine, see my other comment

seeM (approved):
Good catch :)!

nbdev/quarto.py
```diff
@@ -147,8 +147,10 @@ def nbdev_readme(
     _render_readme(cfg_path, path, chk_time)
     if (tmp_doc_path/'README.md').exists():
         _rdm = cfg_path/'README.md'
+        _rdmi = tmp_doc_path/(Path(get_config().readme_nb).stem + '_files')
```
jph00, line 150: You can remove Path()  here because '/' can take str as rhs

nbdev/quarto.py
```diff
@@ -147,8 +147,10 @@ def nbdev_readme(
...
         if _rdm.exists(): _rdm.unlink() # py37 doesn't have arg missing_ok so have to check first
         move(str(tmp_doc_path/'README.md'), cfg_path) # README.md is temporarily in nbs/_docs
+        if _rdmi.exists(): move(str(_rdmi), cfg_path) # Move Supporting files for README
```
jph00, line 153: If move() can't take a str we should modify jt

nbdev/quarto.py
```diff
@@ -259,9 +261,9 @@ def nbdev_quarto(
     pys = globtastic(path, file_glob='*.py', **kwargs).filter(_is_qpy)
     for py in pys: _exec_py(py)
     if preview: os.system(f'cd "{path}" && quarto preview --port {port}')
-    else: _sprun(f'cd "{path}" && quarto render')
-    if not preview:
-        nbdev_readme.__wrapped__(path, doc_path)
+    else: 
+        nbdev_readme.__wrapped__(path, doc_path) # README must be generated BEFORE render because of how index.html is generated by Quarto
```
jph00, line 265: Readme should only be generated if not preview

nbs/01_config.ipynb
```diff
@@ -189,7 +189,16 @@
    "cell_type": "code",
    "execution_count": null,
    "metadata": {},
-   "outputs": [],
+   "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "/Users/hamel/opt/anaconda3/envs/nbdev/lib/python3.9/site-packages/ghapi/core.py:99: UserWarning: Neither GITHUB_TOKEN nor GITHUB_JWT_TOKEN found: running as unauthenticated\n",
```
jph00, line 197: Can you define the env var and rerun so this doesn't appear in docs?

jph00 (changes requested):

nbdev/quarto.py
hamelsmu: This value that is returned isn't a path though

nbdev/quarto.py
hamelsmu: Isn't that what the else statement does here?

nbs/01_config.ipynb
hamelsmu: This cell is hidden, BTW so doesn't appear in the docs!  but I can define it so it doesn't clutter the diff 😃

hamelsmu: @jph00 have some follow up questions here if you can take a look.  I didn't understand a few things.

nbdev/quarto.py
hamelsmu: sure

hamelsmu: @jph00 this isn't our move, this is `move` from shutil, [which doesn't accept Path like objects before Python 3.9](https://docs.python.org/3/library/shutil.html#shutil.move)

BTW - I saw this in the code elsewhere other than this diff, and assumed that this is was for a reason so I followed the convention - looks like it was justified

hamelsmu:
@jph00 have some follow up questions here if you can take a look.  I didn't understand a few things.

nbdev/quarto.py
hamelsmu: I think it clicked with me just now that the LHS `tmp_doc_path` is a path, even though `get_config().readme_nb` is not!  

Sorry about that.  I can change this

hamelsmu: Actually this doesn't work.  I actually just want the stem of the RHS only, just ignore my previous message :)

nbs/01_config.ipynb
jph00: Mainly just didn't want it in the NB. Not a big deal though. (Mind you it's useful to have this set up in your .zshrc anyway!)

nbdev/quarto.py
jph00: Oh yeah absolutely. Now I'm confused as to why it wasn't done that way before!

nbdev/quarto.py
jph00: Oh right I get it now. I didn't read your original carefully enough

nbdev/quarto.py
jph00: Unless I'm missing something, it seems to be ours?

```
from fastcore.shutil import rmtree,move
```

nbs/01_config.ipynb
hamelsmu: Ok makes sense removed this thanks for explaining

nbdev/quarto.py
jph00: After looking again, I do think the `str` should be able to be removed -- try it and see if it works.

hamelsmu: Yeah it works 🎉

seeM:
LGTM!

jph00 (approved):

PR #931, opened 2022-08-25
add `printit` arg to `nbdev_filter` so it can be called with `fname` and still print to stdout

seeM:
Currently, if you call `nbdev_filter --fname foo` from the command-line it will print to stderr because `nbdev_filter` returns the string, and the `nbdev_filter` binary created by setuptools calls `sys.exit(nbdev_filter(...))` which prints to stderr. This is a minor inconvenience when wanting to do something like `nbdev_filter --fname foo | grep output-file` to test its behaviour. cc @hamelsmu @jph00 

The README change is because I forgot to do `nbdev_prepare` on my last PR

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,8 +36,8 @@ practices because tests and documentation are first class.
 - **Publish packages to PyPI and conda** as well as tools to simplify
   package releases. Python best practices are automatically followed,
   for example, only exported objects are included in `__all__`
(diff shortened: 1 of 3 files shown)


seeM:
The original message wasn't very clear: This changes the default behaviour of `nbdev_filter` to always print to stdout, unless `printit=False` is explicitly passed. The intention being that if you want to use `nbdev_filter` in a library, you would pass `printit=False`

Issue #775, opened 2022-08-06
`nbdev_update` creates a new cell, instead of updating the original code


BirkhoffG:
`nbdev_update` does not work as intended. It will create a new code cell, instead of updating the original code.

My use case:

- python_module
    - a.py
    - b.py
    - ...
- nbs
    - a.ipynb
    - b.ipynb
    - ...
 
If I make changes in `a.py`, and run `nbdev_update`, it will create a new code cell in `a.ipynb`, rather than updating code in this notebook.

version: 2.1.1

hamelsmu:
Thanks for reporting this.  I'm getting a different error: the code gets updated into the same cell, but it becomes a markdown cell without any directives.  

@BirkhoffG do you mind sharing a repo or minimal example

BirkhoffG:
Hi @hamelsmu, here is my repo: 
https://github.com/BirkhoffG/counternet_jax

To reproduce the error: 
1. make some changes in `CounterNetTrainingModule` in `cfnet/training_module.py`.  
2. `cd cfnet`
3. `nbdev_update training_module.py`

hamelsmu:
@BirkhoffG I your project is a nbdev v1 repo.  Have you tried migrating this project to nbdev v2 yet?  Here is a migration guide: https://nbdev.fast.ai/migrating.html

Can you please try that and let us know if that helps?

BirkhoffG:
Yes, at the time I reported the bug, I was using the v2 `nbdev`. I manually downgrade to v1 for now as this bug persists.

The `snyc` module in v2 seems to have some problems (e.g., #769), so I downgrade to v1 for now.

BirkhoffG:
> @BirkhoffG I believe your project is a nbdev v1 repo. Have you tried migrating this project to nbdev v2 yet? Here is a migration guide: https://nbdev.fast.ai/migrating.html
> 
> Can you please try that and let us know if that helps?


> Yes, at the time I reported the bug, I was using the v2 `nbdev`. I manually downgrade to v1 for now as this bug persists.
> 
> The `snyc` module in v2 seems to have some problems (e.g., #769), so I downgrade to v1 for now.

To replicate the issue reported, go to the `refactor` branch, which is using `nbdev` v2.

PR #947, opened 2022-08-27
fix #769

### LINKED ISSUE #769: Update All Python Scripts to nbs similar to `nbdev_update_lib` in v1
I am wondering why `nbdev_update` in v2 works differently in v1. In v1, `nbdev_update_lib` will update all modified python scripts, which is very handy. 

However, in v2, to update changes made in the python module, we need to 
1. `cd python_module`
2. `nbdev_update python_file.py`

This is a cumbersome process. Why not just keep the v1 design by just typing the `nbdev_update` in the parent directory?

hamelsmu:
fixes #769 
Also fixes an additional edge case invlolving `__future__`  related to #775 that I did not catch before


nbs/09_API/02_maker.ipynb
```diff
@@ -495,7 +495,7 @@
     "        f.write(f\"# AUTOGENERATED! DO NOT EDIT! File to edit: {self.dest2nb}.\")\n",
     "        if last_future > 0: write_cells(cells[:last_future], self.hdr, f)\n",
     "        if self.parse: f.write(f\"\\n\\n# %% auto 0\\n__all__ = {all_str}\")\n",
-    "        write_cells(cells[last_future:], self.hdr, f, 1 if last_future>0 else 0)\n",
```
hamelsmu, line 498: This line of code where the `offset` was being set was causing some cell indexes to be wrong, and causing the wrong cell in the notebook to be overwritten upon `nbdev_update`.  This was causing some notebooks to be updated incorrectly when they had a `from __future__ ...`

When I removed this offset, everything now seems to be updating correctly.  However, I don't quite understand why this offset was here in the first place, so perhaps @jph00 this is something you might be able to help me understand?

hamelsmu:
@jph00 this PR is ready to review, but perhaps not ready to merge.  I have made a comment on a line of code whose original purpose I don't understand, even though removing this line of code fixes the issue I was having.  

I suggest that we discuss that line of code more before merging because I have a feeling that understanding that context is important

hamelsmu:
It should also be worth noting how I went about testing this.  

1. Without making any changes, running `nbdev_update` from the root doesn't result in any git diffs at all in any notebooks (except for one newline in a single notebook).
2. Made changes in **every single module** (the .py files) and studied the resulting diffs in the notebook(s) to make sure that the notebooks are updating correctly.  I tried making changes in the last cell an also the first cell and then also middle cells.

jph00:
This is such an awesome PR - really love the testing process and description, and double-checking the odd line of code (which appears to be because I changed how things were exported but never changed the index!)

lvgtm (looks *very* good to me)

PR #955, opened 2022-08-29
setup GitHub repo automatically

hamelsmu:
## Enhancements Added

1. Updates the URL and Description
2. Enables GitHub Pages from the root of the `gh-pages` branch (some slighly hacky things are required to make this happen see below comments).  

### Creating a repo with the gh cli
- `gh repo create ntest --clone --public && cd ntest && nbdev_new`
- edit settings.ini and run `nbdev_new` again
- delete settings.ini and run `nbdev_new` again

### Creating a repo with the Github web interface
- navigate to https://github.com/new
- Create a new repo, clone it
- run `nbdev_new`

In each case, I verified that the description, url were updated and that pages are properly enabled. 

cc: @jph00 @seeM


nbdev/cli.py
```diff
@@ -82,13 +82,30 @@ def _render_nb(fn, cfg):
...
 
 # %% ../nbs/09_API/12_cli.ipynb 11
+def _update_repo_meta(cfg):
+    "Enable gh pages and update the homepage and description in your GitHub repo."
+    token=os.getenv('GITHUB_TOKEN')
+    if token: 
+        from ghapi.core import GhApi
+        api = GhApi(owner=cfg.user, repo=cfg.repo, token=token)
+        try: # repo needs something in it before you can enable pages
+            cmds = L(['git add settings.ini', "git commit -m'add settings'", 'git config push.default current', 'git push'])
```
hamelsmu, line 92: When you are first initializing a github repo, it cannot be empty before you enable pages.  It has to contain at least one file.  I don't love executing git commands on behalf of the user.  Another alternative would be to commit an empty file, but that clutters the repo IMO.  

Curious on what your opinions are here.  I put this in its own try block because I suspect something could go wrong here and I don't want to disrupt the onboarding process.

nbdev/cli.py
```diff
@@ -82,13 +82,30 @@ def _render_nb(fn, cfg):
...
+    token=os.getenv('GITHUB_TOKEN')
+    if token: 
+        from ghapi.core import GhApi
+        api = GhApi(owner=cfg.user, repo=cfg.repo, token=token)
+        try: # repo needs something in it before you can enable pages
+            cmds = L(['git add settings.ini', "git commit -m'add settings'", 'git config push.default current', 'git push']) 
+            cmds.map(partial(run, ignore_ex=True))
+            api.enable_pages(branch='gh-pages')
+        except HTTPError: print("Could not enable GitHub Pages automatically.")
+        try: api.repos.update(homepage=f'{cfg.doc_host}/{cfg.doc_baseurl}', description=cfg.description)
```
hamelsmu, line 96: We are able to update the homepage and description with ghapi!

nbdev/cli.py
```diff
@@ -82,13 +82,30 @@ def _render_nb(fn, cfg):
...
+def _update_repo_meta(cfg):
+    "Enable gh pages and update the homepage and description in your GitHub repo."
+    token=os.getenv('GITHUB_TOKEN')
+    if token: 
+        from ghapi.core import GhApi
+        api = GhApi(owner=cfg.user, repo=cfg.repo, token=token)
+        try: # repo needs something in it before you can enable pages
+            cmds = L(['git add settings.ini', "git commit -m'add settings'", 'git config push.default current', 'git push']) 
+            cmds.map(partial(run, ignore_ex=True))
+            api.enable_pages(branch='gh-pages')
```
hamelsmu, line 94: If the repo is empty, we get a `403 error: conflict` if trying to enable pages.  Depending on how you create your repo before initialiing nbdev this is problematic.  This is why the git commit step exists above.

nbdev/cli.py
jph00: I think it's fine - whether we exec git for the user vs use ghapi for the user seems much the same afaict.

FYI I don't think `map` here is useful -- seems a plain `for` loop would be simpler and just as concise. Generally `map` is most useful for the situation where you want to return a list.

jph00: At least until we release it, perhaps best *not* to have the `try`, so we can find out if there are errors, and if so, what errors come up?

PR #959, opened 2022-08-30
Experimental: pre-commit hooks

seeM:
This creates two new hooks for the [pre-commit framework](https://pre-commit.com/): for `nbdev_clean` and `nbdev_export`. They can be used to catch and fix uncleaned and unsynced notebooks, locally, without having to wait for CI. It's also a helpful fallback for people who use editors that we haven't supported yet (e.g. VSCode currently).

It works by defining the `.pre-commit-hooks.yaml` file in our repo root – that's all pre-commit needs! With that, users can easily enable either of these hooks in pre-commit locally.

Here's a rough version of the docs. I haven't included docs in the PR since its still experimental. I'd like for us to try it first then make a call whether to keep and document it. **Note that the hooks will only work once this is merged to master.**

### Install pre-commit

Install pre-commit following [their latest instructions](https://pre-commit.com/#install). Their docs are great so I recommend reading it anyway. Here is a shortened version for convenience.

With pip:

```sh
pip install pre-commit
```

...or with conda:

```sh
conda install -c conda-forge pre-commit
```

...or with homebrew (on mac):

```sh
brew install pre-commit
```

### Configure pre-commit

Create a file named `.pre-commit-config.yaml` in the root of your repo, with the following contents:

```yaml
repos:
- repo: https://github.com/fastai/nbdev
  rev: master # TODO: replace with commit hash or release tag
  hooks:
  - id: nbdev_clean
  - id: nbdev_export
```

Include only the hook(s) you'd like to run, and add any other [supported hooks](https://pre-commit.com/hooks.html).

Install pre-commit hooks into your repo:

```sh
pre-commit install
```

### Make a commit and enjoy pre-commit in action

When you `git commit`, pre-commit will do the following:

1. Stash any unstaged changes
2. Run each hook
    1. If any hook exits with a non-zero status code or changes any files, stop the commit, leaving those changes unstaged. The idea is that you'll stage the changes you want and re-commit
3. Unstash the unstaged changes from earlier. If this would cause any conflicts with changes made by hooks, then those are ignored and the original unstaged changes take preference

It's not as complicated as it might sound. I've found it to be quite seamless in practice. cc @hamelsmu @jph00

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,14 @@
+- &hook
+  id: nbdev_clean
+  name: nbdev_clean
+  entry: nbdev_clean
+  description: "Clean metadata from notebooks to avoid merge conflicts"
+  language: system
+  always_run: true
+  pass_filenames: false
+
+- <<: *hook
+  id: nbdev_export
+  name: nbdev_export
+  entry: nbdev_export
+  description: "Export notebooks to modules and build modidx"


.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,14 @@
+- &hook
+  id: nbdev_clean
```
seeM, line 2: Here's what these mean:

- `language: system` – run `entry` as a shell command. You can also use `language: python` which would install nbdev in its own virtual environment. Since we expect the user to have nbdev installed already, I opted for `system`
- `always_run: true` – always run these hooks regardless of which files were committed. We could also configure so that it only runs when jupyter and/or python files change. I thought it's fast enough that this is a good default
- `pass_filenames: false` – run `entry` as is. If this is `true`, then its called like `nbdev_clean file1 file2 file3` for each staged file. Our commands check all files if run without args, which I thought was a good default, since it matches our current checks in CI

seeM: `language: system` also makes it **much** easier during development, since it will use the current environment's nbdev, fastcore, ghapi, etc

jph00:
Thanks for the detailed and clear PR @seeM ! I think those comments would be a good starting point for docs :)

PR #962, opened 2022-08-30
update tutorial "Check out your docs" section re:pages enablement

hamelsmu:
Since you do not have to enable pages manually, I updated the tutorial to reflect that:

I will explain the workflows of the docs in a separate explanations section that makes show_doc about docs more generally.

image

cc: @seeM @jph00

### DIFF
--- a/nbs/01_Tutorials/01_tutorial.ipynb
+++ b/nbs/01_Tutorials/01_tutorial.ipynb
@@ -464,27 +464,6 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "#### Docs Workflows\n",
-    "\n",
-    "Whenever you change any of the content in your repo, the following workflows will run to generate a docs site (in this order):\n",
-    "\n",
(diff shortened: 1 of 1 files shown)


nbs/01_Tutorials/01_tutorial.ipynb
```diff
@@ -432,67 +432,32 @@
...
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "![](images/nbdev-hello-world-site-initial.png){fig-align=\"center\" .border .rounded .shadow-sm}"
+    "::: {.callout-note}\n",
+    "\n",
+    "nbdev uses GitHub Pages by default because its easily accessible. However, you can use any host you like.  The files for the static site are located in a local `/_docs` directory at the root of your repository after running the command `nbdev_docs` .  This directory is not checked into git and is ignored by `.gitignore`, but you can use these files to deploy to any platform you want.\n",
```
hamelsmu, line 458: I think the part that describes `_docs` is also TMI. I'm going to move that to the explainer section as well in a follow on PR

Issue #964, opened 2022-08-31
`show_doc` includes parsed sections from numpy docstrings


rxavier:
Hi! I'm trying to port an existing codebase to nbdev and running into trouble with docstrings. 

I'm not sure if I'm not following numpy style docstrings  as strictly as I should and tripping over the parser in the process, or whether this is a legitimate bug, so here's a minimal example

I've included a method documented with docments which of course renders nicely.

```
#| export

class Test:
    """A class for testing
    
    Second line.
    
    Parameters
    ----------
    x
        First param
    y
        Second param in which I have added a /
        line break due to length
        
    Attributes
    ----------
    comb
        String combination of `x` and `y`
    """
    def __init__(self, x: str, y: Optional[int] = None):
        self.x = x
        self.y = y
        self.comb = x + str(y)
        
    def print(self,
              end: str = "-", # Line ending
              ) -> None: # No need to return anything, just print
        """Print attrs
        
        Print attrs 2
        """
        print(self.x, self.y, end=end)
```

1. A warning is printed in the rendered docs due to the `Attributes` section (`/Users/rafxavier/Envs/poniard/lib/python3.8/site-packages/fastcore/docscrape.py:225: UserWarning: Unknown section Attributes
  else: warn(msg)`). This is probably because [fastcore/docscrape](https://github.com/fastai/fastcore/blob/79469cd899d4fb51007bb2cc6a7a71e05edbafd6/fastcore/docscrape.py#L97) does not support it. Is this intentional? What about other standard numpy sections like `Raises`? The idea is to document these in notebook markdown?
2. Types are correctly pulled from the signature and descriptions from the docstings, and everything is rendered nicely in a table. However, a `Parameters` section with weird formatting is also included.
<img width="801" alt="image" src="https://user-images.githubusercontent.com/12989861/187760395-f3ec51ff-0227-4332-bd04-c9f9a0822059.png">
3. If the parameter description has a line break, it's not rendered correctly in the table
<img width="783" alt="image" src="https://user-images.githubusercontent.com/12989861/187761416-53bc8b23-5d50-4757-97de-5b0ff9e5b787.png">

seeM:
Thanks for the clear issue write-up! :) #969 fixes (1) redundant sections in numpy docstrings and (2) newlines in the parameters table. Here's what it looks like with the fix:

[...]

PR #969, opened 2022-09-01
fix: `show_doc` includes parsed sections from numpy docstrings

seeM:
Also fix newlines in `DocmentTbl`. cc @hamelsmu @jph00.


seeM:
I forgot to mention: there are example images of the issues in the really nice write-up in #964

hamelsmu:
Yeah that is a nice writeup!

PR #978, opened 2022-09-03
 #967 settings.ini option for choosing preview listen port

tourdownunder:
Created setting for `doc_preview_port` to be used for preview instead of the default 3000.


jph00:
Sorry for the overlap - I fixed this one last night!

PR #979, opened 2022-09-03
put instructions back to enable pages manually

hamelsmu:
@jph00 @seeM IMO the GitHub Actions methodology is extremely flaky for too many use cases and doesn't work well.  In my experience with private repos, even when I was able to sucessfully enable pages via the GitHub api via Actions, GitHub never deployed the site unless I manually went in and changed the settings again, by doing something like changing the path, saving it and changing the path again.  

This is why I'm opening this PR to add the steps back to manually enable GitHub Actions


jph00:
I think we should say this is only necessary for private repos.

hamelsmu:
@jph00 made this change

image

PR #984, opened 2022-09-04
Fix API release docs by moving release.svg and token.png to 09_API/images dir

JungeAlexander:
Fixes nbdev's API release documentation by moving missing `release.svg` and `token.png` from `nbs/images/` to `nbs/09_API/images/`, i.e., this should fix this doc page which is currently broken due to missing image and screencast:

https://github.com/fastai/nbdev/blob/master/nbs/09_API/17_release.ipynb
https://nbdev.fast.ai/09_API/release.html

PS: thanks for a developing this great package!


jph00:
Many thanks!

PR #997, opened 2022-09-07
994 fix spelling

### LINKED ISSUE #994: onwards is misspelled in the config creation
https://github.com/fastai/nbdev/blob/4cd11a3de4d526475170cab158f655e9f855edd0/nbdev/config.py#L76

kevinbird15:
fixes the typo for copywrite and closes #994


nbs/09_API/01_config.ipynb
```diff
@@ -213,7 +213,16 @@
    "cell_type": "code",
    "execution_count": null,
    "metadata": {},
-   "outputs": [],
+   "outputs": [
```
jph00, line 216: Could you please re-run this with a token set up, or just manually remove the error output, so this doesn't appear in the docs?

jph00 (changes requested):
Thanks! Just one minor request

nbs/09_API/01_config.ipynb
kevinbird15: This is done.  I tried to fix it the correct way and get a GITHUB_TOKEN set up, but wasn't able to figure it out.  I'll leave that challenge for another day

PR #1006, opened 2022-09-07
print `nbdev_test` cell errors to stderr instead of using `logging.warning`

### LINKED ISSUE #1004: `nbdev_test` does not restore the original working directory

seeM:
Also:

- fix #1004
- remove non-deterministic `print(duration)`

cc @hamelsmu @jph00


seeM:
@jph00 fixed the merge conflict and CI passes

PR #1027, opened 2022-09-10
`nbdev_new` fails with "Bad credentials" (HTTP 401) if `GITHUB_TOKEN` is set

seeM:
I'm not really happy with the way that I override the token here. Maybe a better way is to update `GhApi` so that if `token=''` it won't default to `$GITHUB_TOKEN`?

I'm also surprised that this causes an issue on public repos, but I verified it with:

```sh
$ GITHUB_TOKEN=foo gh api repos/fastai/nbdev
{
  "message": "Bad credentials",
  "documentation_url": "https://docs.github.com/rest"
}
gh: Bad credentials (HTTP 401)
```

cc @hamelsmu @jph00


jph00:
Sorry I don't understand what problem this is fixing. Why would we expect or want ghapi to work correctly with an invalid token?

seeM:
> Sorry I don't understand what problem this is fixing. Why would we expect or want ghapi to work correctly with an invalid token?

Ohhh good point. I misunderstood the error. I thought that it wasn't working with a **valid** token that didn't have permissions to the fastai org. But I tested now with a valid token that had no repo permissions, and it still works. Closing this

PR #1046, opened 2022-09-13
class method doclinks target `mod.html#method` instead of `mod.html#class.method`

seeM:


seeM:
Thanks to @rxavier for the very helpful minimal repro!

> Okay! Minimal example
> 
> https://github.com/rxavier/test-doclink-method
> 
> https://rxavier.github.io/test-doclink-method/
> 
> You'll see that the hyperlinked `MyClass.method` points to a *section* named *method*, and not the *method show_doc*.

Issue #1050, opened 2022-09-13
`nbdev_export` without adding versions to `__init__.py`


MichaelJFishmanBA:
I'd like to stop `add_init` (via `nbdev_export` via `nbdev_prepare`) from adding the version to `__init__.py`.

Context: The team I work with uses Jenkins to package our libraries, and use automatic versioning. I don't want to manually set the version in `settings.ini` in parallel to our automatic versioning, and by default `add_init` uses version `0.0.1` if `settings.ini` does not specify a version.

I'm thinking we could make another variable in `settings.ini` called `put_version_in_init` that defaults to `True` so that

1. When `True`, we set the version in `__init__.py` as we currently do
2. When `False`, we skip setting the version in `__init__.py`

I'd be happy to write a PR for this if you think it sounds reasonable.

MichaelJFishmanBA:
[Here](https://github.com/fastai/nbdev/pull/1051) is a PR that does this.

PR #1051, opened 2022-09-13
Setting `put_version_in_init ` to make adding `__version__` to `__init__.py` optional

MichaelJFishmanBA:
Addresses #1050

Adds an option `settings.ini` called `put_version_in_init` that defaults to `True` so that

1. When `True`, we set the version in `__init__.py` as we currently do
2. When `False`, we skip setting the version in `__init__.py`


jph00:
lgtm. FYI @seeM just simplified how types work, so there's a conflict now. Do you mind resolving the conflict so I can merge this?

jph00:
Actually might be easier for me to just make the change directly at my end -- I'll close this and note your contribution in the changelog.

PR #1057, opened 2022-09-14
set `output-file` using notebook filename instead of `default_exp`

### LINKED ISSUE #1032: Docs inconsistant for nbs and source code with sub-folders
When the source code contains sub-folders, for example in my repo I have:

https://github.com/MannLabs/alphapeptdeep/blob/main/nbdev_nbs/rescore/percolator.ipynb (rescore is the sub-folder in both nbs and py module).

The html path the of this ipynb will be `.../rescore/rescore.percolator.html` (see https://mannlabs.github.io/alphapeptdeep/rescore/rescore.percolator.html), which is fine.

However, in my [index.ipynb](https://github.com/MannLabs/alphapeptdeep/blob/main/nbdev_nbs/index.ipynb) (see https://mannlabs.github.io/alphapeptdeep/#dda-rescoring), I called:

```
from peptdeep.rescore.percolator import Percolator
```
The \`Percolator\` term in the markdown cell will be translated to `https://mannlabs.github.io/alphapeptdeep/rescore/percolator.html#percolator`, which is inconsistant with `.../rescore/rescore.percolator.html#percolator`.

The nbdev version is 2.3.0 from master branch.

seeM:
- fix #1032


jph00:
Good thinking!

PR #1058, opened 2022-09-14
fix `nbdev_update`

seeM:
Apologies for the huge PR! I'm happy to break it down into smaller PRs if you prefer

- fix bug in `nbdev.maker.update_import` which meant that `nbdev_update` didn't convert relative imports without `None` module (e.g `from . import foo` -> `from pkg import foo`)
- fix `FileNotFoundError` in `nbdev_update` by passing the correct py module and corresponding notebook paths
  - refactor `nbdev.doclinks._get_modidx` to use extracted `_iter_py_cells` function to support this
- fix `nbdev_update` introducing whitespace changes to notebooks
  - refactor `nbdev.process.extract_directives` to use extracted `_partition_cell` function to support this


nbdev/maker.py
```diff
@@ -149,7 +149,7 @@ def update_import(source, tree, libname, f=relative_import):
         nmod = f(imp.module, libname, imp.level)
         lin = imp.lineno-1
         sec = src[lin][imp.col_offset:imp.end_col_offset]
-        newsec = re.sub(f"(from +){'.'*imp.level}{imp.module}", fr"\1{nmod}", sec)
+        newsec = re.sub(f"(from +){'.'*imp.level}{imp.module or ''}", fr"\1{nmod}", sec)
```
seeM, line 152: This is to fix:

> fix bug in `nbdev.maker.update_import` which meant that `nbdev_update` didn't convert relative imports without `None` module (e.g `from . import foo` -> `from pkg import foo`)

nbdev/sync.py
```diff
@@ -8,7 +8,9 @@
 from .config import *
```
seeM, line 8: I tested this by running `nbdev_update` in the nbdev repo and ensuring that it didn't change any of our notebooks

nbdev/doclinks.py
```diff
@@ -43,35 +43,43 @@ def patch_name(o):
     return _sym_nm(a,o)
 
 # %% ../nbs/api/doclinks.ipynb 9
-def _nbpath2html(p): return p.with_name(re.sub(r'\d+[a-zA-Z0-9]*_', '', p.name.lower())).with_suffix('.html')
+def _iter_py_cells(p):
```
jph00, line 46: Nice approach!

nbdev/sync.py
jph00: Good idea

jph00:
Looks great!

PR #1062, opened 2022-09-15
use migration commands that work with default macos shell

### LINKED ISSUE #1060: MacOS problems on migration from nbdev1
The migration recipe relies on bash and globstar as a shell option.  MacOS defaults to zsh (easy fix as bash is there still), but it does not include globstar in the build of bash

Certainly my builtin bash doesn't have globstar (MacOS 10.15.7 has bash 3.2), and [it appears that globstar was added to bash in v4.0](https://www.linuxjournal.com/content/globstar-new-bash-globbing-option).  Even [very recent builds on the latest hardware](https://kizu514.com/blog/install-upgraded-gnu-bash-on-a-macbook-pro/) are still bash v3. 

Apparently [Apple doesn't want to update to later versions of bash to avoid GPL issues](https://dev.to/emcain/how-to-change-bash-versions-on-mac-with-homebrew-20o3).

I can think of two separate solutions
- the migration script uses an alternative to globstar (e.g., find)
- OR include directions that MacOS users need to install new bash using homebrew (as suggested in the links above) and must switch to bash for executing the globstar perl scripts

seeM:
- fix #1060


hamelsmu:
Note: there is some overlap b/w this and the migration script which also migrates directives.  But I don't think it matters.  Just wanted to observe or share that because I even sometimes miss all the moving parts

PR #1079, opened 2022-09-16
respect `#|hide` and `#|include: false` for showdoc

### LINKED ISSUE #1076: Export class to library but hide from documentation
Is it possible to use both hide and export directions such that my cell is exported to the libraries but is not shown in the documentation? 

My particular case: I am creating a class through multiple cells, e.g.:

Cell 1:
```
#| export 
class foo():
    def __init__():
        ''' Description of the class, which is shown in the docs.'''
```

Cell 2 (I want this one to be exported but hidden in the doc):
```
#| hide (?)
#| export (?)

class foo(foo):
    def func():
        ''' Description of the function, which I want to later show in the doc with show_doc(foo.func). '''
```

Combining both directives currently does not work, and in the library that cell shows again the class definition. I want to hide it and use later show_doc to show only the foo.func function I created. Is this possible? If not, is there a preferable way of defining classes over cells in nbdev?

hamelsmu:
closes #1076 

I tested this extensively with both `#|hide` and `#|include: false` to make sure the code is still exported but the show doc is not called

`#|include: false` is the same thing as `#|hide`

@jph00 @seeM


nbdev/processors.py
```diff
@@ -59,8 +59,10 @@ def _show_docs(trees):
     return [t for t in trees if isinstance(t,ast.Expr) and nested_attr(t, 'value.func.id')=='show_doc']
 
 def cell_lang(cell): return nested_attr(cell, 'metadata.language', 'python')
+
+def _want_hide(c): return 'hide' in c.directives_ or c.directives_.get('include:') == ['false']
```
hamelsmu, line 63: I found it odd that the colon is included in the key `include:` vs `include`.  Looking at the code this seems deliberate and is even in the tests for `processor.extract_directives`.

jph00: I'm not sure why it wouldn't include the `:`? That's the thing that let's you know it's a key:value thing.

BTW why did you abstract out `_want_hide` here? Seems like you could easily inline it since you're only using it in one place?

jph00 (approved):
Feel free to merge when ready - just one question inline

nbdev/processors.py
hamelsmu: Would you mind taking a look at the new way I just pushed?  Just want to check if that is more in line with what you would do

PR #1084, opened 2022-09-16
`nbdev_new` queries `branch` from GitHub

### LINKED ISSUE #1080: `nbdev_new` queries `branch` from GitHub
The default branch names used by GitHub and nbdev differ:

- GitHub [uses "main" as the default branch name for new repositories](https://github.com/github/renaming#new-repositories-use-main-as-the-default-branch-name)
- nbdev [uses "master" for the default branch name](https://github.com/fastai/nbdev/blob/master/nbdev/config.py#L41)
 
This causes nbdev commands to fail unless `settings.ini` is updated.

seeM:
- fix #1080
- `branch` is now a required setting (it has no default and must be specified in your `settings.ini`)


hamelsmu:
fixed minor conflict for you

seeM:
@hamelsmu legend! 😄

PR #1090, opened 2022-09-18
restructure "Notebook best practices" tutorial

seeM:
- move example back to the end
- add preview card
- extract "Why nbdev" explanation from intro

@hamelsmu @jph00 any suggestions/hesitations? @hamelsmu you originally wrote most of "why nbdev" so you might have a better idea for how it'd look as its own page (and whether that title is accurate), although that all could be made in a follow-up PR


jph00:
Great!

(The "why nbdev" page isn't great yet, because we really do need that "why notebooks" page for it to make sense!)

Issue #13, opened 2019-12-04
nbdev.export.relative_import does not convert all possible import syntax


ddobrinskiy:
https://github.com/fastai/nbdev/blob/242331c4f05ce6e275dcbacaea7915545e29bbd5/nbdev/export.py#L182

Given this cell

```python
# export

import my_package.uploader

import my_package.utils as u

from my_package.test import *
```
**Expected behavior:** converts to  

```python
import .uploader

import .utils as u

from .test import *
```

**Observed behavior:** converts to 

```python
import my_package.uploader

import my_package.utils as u

from .test import *
```

sgugger:
Good catch! Since we only use the last syntax, we didn't think to include the first. Should be fixed now.

quocpeyrot:
I'm getting a syntax error with the form "import .utils as u"

from the documentation it seems this is indeed illegal:
https://docs.python.org/3/reference/import.html

`Absolute imports may use either the import <> or from <> import <> syntax, but relative imports may only use the second form;`

from that documentation, it seems relative import should only be used with the form `from [...] import [...]`

sgugger:
It would be good to document this, if you want to suggest a PR in that direction.

quocpeyrot:
I'm not sure I understand what you're asking.

From what I can tell from the documentation (and the fact that I'm getting an SyntaxError with Python 3.7), the request from this issue was bogus, and should be reverted?

It doesn't seem to work with Python 3.7, and the documentation from 3.8.1 linked above tells the same.

sgugger:
The first one is still useful (a simple import lib_name.module should be changed to import .module). The second one doesn't work indeed, so it should be documented (in this case, since it doesn't work, we don't really care about the conversion or am I missing something?).

quocpeyrot:
I'm confused, are you saying that `import .module` is valid?
It doesn't seem to be working for me either (and the documentation doesnt tell it either)

(and yes, I think those cases should just be ignored and not converted)

> % python
> Python 3.7.5 (default, Oct 25 2019, 10:52:18)
> [Clang 4.0.1 (tags/RELEASE_401/final)] :: Anaconda, Inc. on darwin
> Type "help", "copyright", "credits" or "license" for more information.
> >>> import .foo
>   File "<stdin>", line 1
>     import .foo
>            ^
> SyntaxError: invalid syntax
> >>> import foo
> >>> foo.foo()
> 'foo'
> >>>
>

sgugger:
[...]

PR #1094, opened 2022-09-18
fix #1091: `nbdev_new` pins on major+minor version of `nbdev-template`

### LINKED ISSUE #1091: `nbdev_new` pins on major+minor version of `nbdev-template`
Since every version of nbdev uses latest nbdev-template, we can't make any breaking changes in nbdev-template. I propose that `nbdev_new` pins the major+minor version of nbdev-template and that nbdev-template bumps minor version on breaking changes.

### Example

Implementing https://github.com/fastai/nbdev/issues/1083 would be simpler with this change.

Currently, we can implement it by updating the default in `_apply_defaults` and having `nbdev_new` move a known list of files (`*.ipynb`, `*.css`) to `cfg.nbs_path`.

We wouldn't have to keep that list in our code if we could move those files into a new sub-dir of `nbdev-template`, but that would break `nbdev_new` for prior versions of nbdev.

cc @hamelsmu @jph00

seeM:
- fix #1091


nbs/api/cli.ipynb
```diff
@@ -172,6 +172,25 @@
     "                  \"Use a token with the correction permissions or perform these steps manually.\")"
```
hamelsmu, line 172: <pre class="ql-syntax" spellcheck="false">Line #13.            tag = _nbdev_template_release((1,1)).tag_name
</pre><p>I wonder if there is a way to make it such that we can remind ourselves that we must change this line of code when we make a breaking change in the template repository.  Surely, we will forget to this otherwise?  </p><p> </p><p>Any ideas?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/fastai/nbdev/pull/1094/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='12'/>

seeM: Good point. I don't have a good answer yet, best I can think of is an e2e test for `nbdev_new` that (hopefully) fails if we forget to bump

seeM: @jph00 I'm curious to hear your take on this

seeM:
Closing. https://github.com/fastai/nbdev/issues/1103 is a better solution to the underlying problem

PR #1102, opened 2022-09-20
The Divio system is called Diátaxis these days

simonw:


simonw:
The author of that framework moved on from Divio and now works at Canonical - he's rebranded the documentation system Diátaxis and is promoting that term and continuing to develop it.

jph00:
Thank you for your 1st nbdev commit! :D

PR #1113, opened 2022-09-23
`exec_doc` supports re-rendering widgets

seeM:
If users update widgets with the `exec_doc` directive, the widget "view" is updated (in the cell output), but the old widget "state" is used (in notebook metadata). This refreshes widget state using `ipywidgets.Widget.get_manager_state`.

See [Tyler's comment](https://forums.fast.ai/t/approach-for-re-executing-notebooks-on-a-schedule/99227/3?u=seem) for context. cc @tylere


seeM:
Hmm I just realised that this always overwrites existing widget state (e.g. if a user saved it in their IDE). So if your `exec_show_docs` doesn't render any new widgets, all of your existing widgets will break...

seeM:
One solution would be to merge the `exec_show_docs` state dict with any existing one in the `nb`, since the state dict is keyed on widget UID.. Will update the PR with that

hamelsmu:
Do you think it might make sense to add a test for this?

hamelsmu (approved):

PR #1122, opened 2022-09-27
`black_formatting` setting is ignored

jmoralez:
The `black_formatting` option is now calling `str2bool` from fastcore so it's value when retrieved by `cfg.get('black_formatting')` is either 0 or 1, so the code path that does the black formatting is currently inaccessible. This changes the comparison to check against `1`.

### DIFF
--- a/nbdev/export.py
+++ b/nbdev/export.py
@@ -32,7 +32,7 @@ def black_format(cell, # Cell to format
     "Processor to format code with `black`"
     try: cfg = get_config()
     except FileNotFoundError: return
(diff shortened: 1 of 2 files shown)


seeM:
Thanks for the contribution! 😄 This can be simplified a little since 1 is "truthy" and 0 is "falsy". We also don't need to provide the default here anymore, they're now defined in `_apply_defaults` [here](https://github.com/fastai/nbdev/blob/5de094d89d249113ef022d96478cbadb4784429c/nbdev/config.py#L58). So you could do:

```python
if (not cfg.black_formatting and not force) or cell.cell_type != 'code': return
```

jmoralez:
Thanks for the suggestion! Applied in 19a95dd

PR #1133, opened 2022-09-30
Fix pathological regex case

### LINKED ISSUE #1132: Clean takes forever on notebooks with large output
`nbdev_clean` takes multiple minutes (I didn't wait for it to finish) on this notebook: https://github.com/dleen/nbdev/blob/repro/plotly2.ipynb

The exact output that trips it up is: https://github.com/dleen/nbdev/blob/repro/pathological.txt

The problem is the regex that searches for ids is slow on large input: https://github.com/fastai/nbdev/blob/master/nbdev/clean.py#L48

I believe it gets hung up on this as there are ~8,000 occurrences of `<` since it is javascript in the output.

My first solution is to substring search for `at 0x` which seems to be the key part of the regex. Don't bother doing the regex check if that substring isn't found as the regex won't match anything. Worst case the search complexity is some small multiple of the input: https://stackoverflow.com/a/35220792.

dleen:
fixes #1132


dleen:
@seeM @jph00

jph00:
Good idea

PR #1135, opened 2022-09-30
Write to stdout

### LINKED ISSUE #1134: Write to stdout is broken for nbdev_clean
It calls execnb `write_nb` with a stream but execnb only accepts paths

dleen:
fixes #1134

@jph00


jph00:
thank you!

PR #1136, opened 2022-09-30
Don't use generics - breaks on Python 3.8

dleen:
cc @jph00


jph00:
Oh thanks - yeah I hate generics in Python anyway so I prefer that anyway!

PR #1137, opened 2022-09-30
Try latest quarto release

dleen:
See if this fixes the build


dleen:
Seemed to work... assume 165 was a buggy release @jph00

jph00:
Yup I emailed JJ about it and he fixed it overnight, so I've updated to just use the latest again.

Issue #1140, opened 2022-09-30
Images in index.ipynb causing deployments issues


KGrewal1:
Images in the index seem to cause the error
 File exists: '/home/runner/work/hypersheets/hypersheets/index_files'

when running Deploy to GitHub Pages

nueshym:
Maybe your issue is the same as mine fastai/nbdev#1106 if the error looks like:
```
FileExistsError: [Errno 17] File exists: '<abs_path_to_root>/index_files'
```

PR #1143, opened 2022-10-03
fix copying of index assets

### LINKED ISSUE #1106: nbdev_readme() fails on the second run for the notebook with support files (e.g. Fig image).
Minimal setup to reproduce this error:
https://github.com/nueshym/nbdev_readme_error/blob/main/nbdev_readme_test.ipynb

# Problem

Executing `nbdev_readme()` while there's already exported support files folder exists at root directory cause:
```
FileExistsError: [Errno 17] File exists: '<abs_path_to_root>/index_files'
```

At https://github.com/fastai/nbdev/blob/8cf8d257ae1ba24bacb5f5a82910772959c5bc4f/nbdev/quarto.py#L217

# Expected behaviour

README.md and support files are exported to root directory.

# Actual behaviour

support files are not exported.

# Solution

Check if the directory exists before `copytree`.

# ...

And there's another weird bug closely related to this one:
You can't use `readme.ipynb` for `readme_nb`.
failed with:
```
ERROR: NotFound: No such file or directory (os error 2)
```
Seems the name `readme` conflicting with something at quarto or pandoc level.
I will submit as an another issue if you want but that requires another fresh project to reproduce...

Thanks for great project!

hamelsmu:
fixes #1106 
Also fixes https://github.com/fastai/nbdev/issues/1140

### DIFF
--- a/nbdev/quarto.py
+++ b/nbdev/quarto.py
@@ -213,7 +213,7 @@ def nbdev_readme(
         _rdmi = tmp_doc_path/(idx_cache.stem + '_files')
         if readme_path.exists(): readme_path.unlink() # py37 doesn't have `missing_ok`
         move(readme, cfg_path)
-        if _rdmi.exists(): copytree(_rdmi, cfg_path/_rdmi.name) # Move Supporting files for README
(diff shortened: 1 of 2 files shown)


jph00:
Thanks! :)

PR #1157, opened 2022-10-13
Experimental: Users can provide extra processors via the `procs` key in `settings.ini`

seeM:
Add a space-separated list of import paths to your `settings.ini` using the same format as for `console_scripts` but without the name. For example:

```ini
procs = nbdev_extensions.codenotes:NoteExportProc
```

cc @muellerzr


jph00:
Interesting!

Is the CI failure due to some Quarto issue?

seeM:
> Is the CI failure due to some Quarto issue?

Seems so. We have some cells with both `#| echo` and `show_doc`. The `clean_show_docs` processor then adds another `#| echo: false` leading to the quarto error: `ERROR: YAMLException: duplicated mapping key`.

I don't think we changed anything here so I guess quarto added stricter parsing. Should we make `clean_show_doc` only add `#| echo: false` if it isn't already set?

jph00:
Yes I think we should

PR #1161, opened 2022-10-14
Add note about unions

### LINKED ISSUE #1145: Best practice in type annotation
From the [tutorial](https://nbdev.fast.ai/tutorials/best_practices.html), it seems that `nbdev` favors simple type annotations. For example, it uses `int|str`, rather than `Union[int,str]`. 

Albeit simplicity (which I do enjoy), this type of annotation (i.e., `|`) is only introduced after Python 3.10. However, `nbdev`' does support Python>=3.7. How would one who uses Python 3.7 - 3.9 annotate `Union`?

dleen:
Fixes #1145


seeM:
Thank you!

PR #1164, opened 2022-10-17
fix #1163

### LINKED ISSUE #1163: Setting `#| echo` in a cell with `show_doc` causes a Quarto error
If a cell has both `#| echo` and `show_doc`. The `clean_show_docs` processor then adds another `#| echo: false` leading to the quarto error: `ERROR: YAMLException: duplicated mapping key`.

We can fix this by making `clean_show_doc` only add `#| echo: false` if there isn't already an `echo` directive in the cell.

_Originally posted by @seeM in https://github.com/fastai/nbdev/issues/1157#issuecomment-1279943199_

seeM:
fix #1163


jph00:
Thank you for fixing this!

PR #1165, opened 2022-10-17
fix duplicated sections in the sidebar

seeM:
Tbh I'm not really sure why it works, but it does.

Before:

image

After:

image


jph00:
Thanks. I think it used to work. Perhaps something changed in Quarto...

PR #1170, opened 2022-10-19
Option to restrict generated lib line width

itepifanio:
This MR adds the `lib_max_width` option to nbdev. This option constrains the generated lib to specific line width.

When running some lint on the nbdev generated library sometimes the `__all__` function it's too big. This MR avoids lint warnings by adding the `lib_max_width` option to restrict the size of `__all__` lines.


hamelsmu:
What linter are you using?  Can you set your linter to ignore `__all__` lines ?  

For the rest of the generated code, if you are worried about width, you can use the `black_formatting=True` option in `settings.ini` that will format your code according to black.  

Perhaps that could solve your issues?

itepifanio:
@hamelsmu sorry for the late reply. I'm using [flake8](https://flake8.pycqa.org/en/latest/), I couldn't find a way to ignore the `__all__` lines but the `black_formatting=True` worked. Thanks!

PR #1172, opened 2022-10-22
Let proc not have code

muellerzr:
I'm writing a lazy processing system that let's me write quarto directives without needing `:::`, so rather than doing:

```markdown
::: {.column-margin}
Some column stuff
:::
```
I can do 
```markdown
#| layout .column-margin
Some things
```

And with the current setup if I wanted to do something such as start a panel tabset group that will go for a while, it just needs to contain a markdown cell with the *start* of the groups, so my logic operates such that my entire cell is:

```markdown
#| layout .panel-tabset group='setup' start
```
And then later I have
```markdown
#| layout end
```
With no code.

### DIFF
--- a/nbdev/process.py
+++ b/nbdev/process.py
@@ -58,6 +58,7 @@ def _partition_cell(cell, lang):
     if not cell.source: return [],[]
     lines = cell.source.splitlines(True)
     first_code = first_code_ln(lines, lang=lang)
+    if not first_code: return lines,[]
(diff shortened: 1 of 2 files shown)


jph00:
Closed already?... Wazzup?

muellerzr:
@jph00 already supported actually, I was fighting non-existent bugs by not using the latest quarto version I think :) if I come across the need for this again in a better way for my head to think about it I'll reopen the PR with the modified changes

PR #1174, opened 2022-10-23
Fix nbdev pre-commit hook

johnnv1:
By fixing the `language` at the pre-commit hook config, this remove the necessity to manually install `nbdev` with pre-commit. 

Before this fix, when using the pre-commit, we have a duplicated `nbdev` environments, one at the pre-commit files (which should be the one used) and another in your environments (which was what was being used). With this fix, the pre-commit will use the files from pre-commit environment -- which will be cached between multiples repo.

Also, this will possibly fix the `nbdev` hook to work with [pre-commit ci](https://pre-commit.ci/).

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -3,7 +3,7 @@
   name: nbdev_clean
   entry: nbdev_clean
   description: "Clean metadata from notebooks to avoid merge conflicts"
-  language: system
+  language: python
   always_run: true
   pass_filenames: false
 
--- a/nbs/tutorials/pre_commit.ipynb
+++ b/nbs/tutorials/pre_commit.ipynb
@@ -43,52 +43,7 @@
    "id": "0c3c7192-892a-4ff9-acb0-a89fe8a2e4bc",
    "metadata": {},
    "source": [
-    "## Install nbdev and pre-commit"
-   ]
-  },
-  {
-   "cell_type": "markdown",
-   "id": "29cade2a-32ea-4640-8585-ef0448f8b7f3",
-   "metadata": {},
-   "source": [
-    "To get started, install nbdev:"
-   ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "id": "3acdaca2-0fa0-4355-a585-e12174b7a304",
-   "metadata": {},
-   "outputs": [
-    {
-     "name": "stdout",
-     "output_type": "stream",
-     "text": [
-      "::: {.panel-tabset}\n",
-      "\n",
-      "#### pip\n",
-      "\n",
-      "```sh\n",
-      "pip install -U nbdev\n",
-      "```\n",
-      "\n",
-      "#### conda\n",
-      "\n",
-      "```sh\n",
-      "conda install -c fastai nbdev\n",
-      "```\n",
-      "\n",
-      "\n",
-      ":::\n",
-      "\n",
-      "\n"
-     ]
-    }
-   ],
-   "source": [
-    "#| echo: false\n",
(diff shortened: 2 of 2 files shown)


johnnv1:
Pinging @seeM or @jph00 to reviews this

seeM:
Thanks for the PR! Could you please elaborate on how `language: system` leads to duplicated nbdev environments? I originally chose it because we expect users to install nbdev when working with nbdev packages, and I thought it would be confusing if a different environment was setup just for pre-commit.

johnnv1:
For example, if i want to add nbdev to any repo, the `.pre-commit-config.yaml` will be something like this:
```yml
...
- repo: https://github.com/fastai/nbdev/
  rev: xx
  hooks:
    - id: nbdev_clean
```

with just this information, pre-commit will know, from where and what he needs to setup and run the hook...

In a simplistic manner, we can say, pre-commit will create an env for that repo in your machine (on linux at `~/.cache/pre-commit/
`) with all hooks from the `.pre-commit-config.yaml`. So, as nbdev is a python library, pre-commit will run the hook with python (if `language: python` -- which is located inside the pre-commit environment for that repository).


-----
if we have `language: system` the pre-commit will look for the command into the system path -- which will cause the user to need the nbdev available on it -- so the nbdev installed inside the pre-commit environment will be ignored... basically this is an antipattern (wrong use of this setting `.pre-commit-hooks.yaml`) of the pre-commit tool

If you want to user have the nbdev installed into the environment, can just config the `.pre-commit-config.yaml` as something like this -- but this is not recommended/seems to be a bad use of the tool:
```yml
...
 - repo: local
  hooks:
    - id: nbdev-clean
      name: nbdev-clean
      entry: nbdev_clean
      language: system
      pass_filenames: false
      args: ['--fname', './']
```

----
to be able to use nbdev only as a hook now, working on pre-commit.ci, it is currently necessary to do this workaround: https://github.com/lapix-ufsc/dl-tutorials/blob/cf9ad66d95756e36c233b9762c755e01620a013c/.pre-commit-config.yaml#L24

But the idea of having a `.pre-commit-hooks.yaml` is to users don't need to do this workaround above, and can directly use the tool as a hook

johnnv1:
ah, additionally, pre-commit simplifies it, so the user doesn't need to know all this, and it just "magically works" ... so that it does not need to have several dependencies for the development environment because the pre-commit will take care of this

seeM:
Thanks so much for the clear and detailed response @johnnv1! Is it not possible to install your own dependencies in pre-commit CI instead of reconfiguring the hook?

The reason I'm hesitant to use `language: python` is that we expect the user to also have nbdev installed in their own environment (for example, to run things like `nbdev_docs`, `nbdev_preview`, `nbdev_install_hooks`, `nbdev_release_*`, and so on). I suspect having nbdev installed separately in a user's environment and in pre-commit's environment might lead to using different versions which could cause confusion. What do you think?

johnnv1:
For me, I don't see it this way, I prefer the `.pre-commit-hooks.yaml` configured in the right way...
Like, for example, I just want to use `nbdev` clean hook, into a repo, for that i don't want to have to install 2x the nbdev... I'm able to use the nbdev at the pre-commit.ci, but it's a workaround because of this wrong config... 

Seem more clear, if you want/prefer the user to use the same nbdev that is in the development env, it makes more sense to configure the `.pre-commit-config.yaml` as something like this -- but this is not recommended/seems to be a bad use of the tool:
```yml
...
 - repo: local
  hooks:
    - id: nbdev-clean
      name: nbdev-clean
      entry: nbdev_clean
      language: system
      pass_filenames: false
      args: ['--fname', './']
```

Issue #1190, opened 2022-10-30
Notebooks and library not in sync (even though they are)


omlnaut:
In my project I can't get the build-pipeline to work again. It fails with the error, that my notebooks and the library are not in sync.
I already tried:
- upgrading local nbdev (is 2.3.7.)
- running nbdev_export locally, pushing the changes (there were none)
- deleting the library dir locally, pushing that. Then run nbdev_export locally to re-generate the lib folder, pushing that
- cloning the repository into a new local folder and running nbdev_export there. Again, there are no changes when running git status
- Deleting the github-runner-cache
- Performed the exact steps done by https://github.com/actions/checkout on my local machine. Ran `nbdev_export`. Still No changes shown by git status.

To no avail. It seems that whatever version the gitlab runner pulls is different than what I see in the repo.

Here is the pipeline in question:

https://github.com/omlnaut/pymoq/actions/runs/3356523622/jobs/5561653458

Any help is appreciated :)

Update:
I figured out how to ssh into the github runner while it's executing the workflow. The issue is not that the repository is checked out incorrectly. Instead, it seems that `nbdev_export` doesn't handle the case that I have one notebook (04_...) with a `default_export` and another notebook (06_....) that has specific cells exporting to the same target. On my local machine this works as intended, but when I run `nbdev_export` via ssh on the runner, only the code from the 04_... notebook gets copied to the target file.

omlnaut:
Just now I tried cloning the repository into a new local folder and running nbdev_export there. Again, there are no changes when running `git status`.

omlnaut:
Also tried deleting the github-runner-cache, error persists

omlnaut:
I think I figured out the error here.

The notebook that generates the initial `mocking/functions.py` file (via `default_exp`) has to be exported before the notebook that adds to `mocking/functions.py` via individual `export mocking.funcions` tags.
`nbdev_export` uses globtastic to get a list of notebook paths, then iterates over the list, exporting them in the order they were returned by globtastic. It seems that this order is different on my local machine than on the github runner.

I fixed the issue for now by explicitly adding a `nbdev_export --path nbs/06_verify.ipynb` call right below the `nbdev_export` call in `test.yaml`. This fixes the issue for now.

For a longterm fix I think it might be useful to sort the notebook-paths alphabetically ascending by filename. This would make the export order deterministic (and probably in line with what one would expect when naming the notebooks 00_.. 01_.... and such.

[...]

PR #1192, opened 2022-10-31
nb_export now exports notebooks in ascending-filename-order

omlnaut:
Fixes the issue described here:
https://github.com/fastai/nbdev/issues/1190

It matters in which order notebooks are exported, as a notebook with #| default_exp A needs to be exported before any notebooks than contain a cell starting with #| export A. The order in which the notebooks are exported so far is arbitrary (https://docs.python.org/3/library/os.html#os.listdir).

This PR adds an option to nbglob to sort the results by an arbitrary key-function. Then it extends nbdev_exports call to nbglob by passing a sorting-key-function that sorts by the notebook filenames.

The pipeline linked in the issue above is fixed when using the nbdev fork that is requested to be merged here.

### DIFF
--- a/nbdev/doclinks.py
+++ b/nbdev/doclinks.py
@@ -100,15 +100,18 @@ def _build_modidx(dest=None, nbs_path=None, skip_exists=False):
 
 # %% ../nbs/api/doclinks.ipynb 20
 @delegates(globtastic)
-def nbglob(path=None, skip_folder_re = '^[_.]', file_glob='*.ipynb', skip_file_re='^[_.]', key='nbs_path', as_path=False, **kwargs):
+def nbglob(path=None, skip_folder_re = '^[_.]', file_glob='*.ipynb', skip_file_re='^[_.]', key='nbs_path', as_path=False, sort_by=None, **kwargs):
     "Find all files in a directory matching an extension given a config key."
     path = Path(path or get_config()[key])
     recursive=get_config().recursive
     res = globtastic(path, file_glob=file_glob, skip_folder_re=skip_folder_re,
                      skip_file_re=skip_file_re, recursive=recursive, **kwargs)
-    return res.map(Path) if as_path else res
+    res = res.map(Path) if as_path else res
+    if sort_by is not None:
+        res.sort(key=sort_by)
+    return res
 
-# %% ../nbs/api/doclinks.ipynb 21
+# %% ../nbs/api/doclinks.ipynb 23
 def nbglob_cli(
     path:str=None, # Path to notebooks
     symlinks:bool=False, # Follow symlinks?
@@ -122,7 +125,7 @@ def nbglob_cli(
     return nbglob(path, symlinks=symlinks, file_glob=file_glob, file_re=file_re, folder_re=folder_re,
                   skip_file_glob=skip_file_glob, skip_file_re=skip_file_re, skip_folder_re=skip_folder_re)
 
-# %% ../nbs/api/doclinks.ipynb 22
+# %% ../nbs/api/doclinks.ipynb 24
 @call_parse
 @delegates(nbglob_cli)
 def nbdev_export(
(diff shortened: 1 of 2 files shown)


seeM:
Good idea! I think this could be brought into `fastcore`'s `globtastic` in future, but seems reasonable to get the fix into nbdev for now. What do you think @jph00 @hamelsmu?

omlnaut:
@seeM @jph00 @hamelsmu 
Any updates on when this can be included in nbdev?

hamelsmu:
I think I agree with you @seeM .  However, I think it would be nice to find a way where the export is order-independent.  It's not 100% clear to me that we have to force things to be this way

nbdev/doclinks.py
```diff
@@ -122,24 +125,24 @@ def nbglob_cli(
...
+# %% ../nbs/api/doclinks.ipynb 24
 @call_parse
 @delegates(nbglob_cli)
 def nbdev_export(
     path:str=None, # Path or filename
     **kwargs):
     "Export notebooks in `path` to Python modules"
     if os.environ.get('IN_TEST',0): return
-    files = nbglob(path=path, **kwargs)
+    files = nbglob(path=path, sort_by=lambda path_str: Path(path_str).name, **kwargs)
```
jph00, line 136:
```suggestion
    files = nbglob(path=path, as_path=True, **kwargs).sorted('name')
```

jph00 (changes requested):
Thanks for the PR! I'm not sure we need to add anything to `nbglob`, since we can just use `sorted` at the calling site as shown.

Not needed in this PR, but we should consider either whether we can remove the need for this sorting entirely, or else look at the `order` frontmatter as well when sorting.

omlnaut:
> Thanks for the PR! I'm not sure we need to add anything to `nbglob`, since we can just use `sorted` at the calling site as shown.
> 
> Not needed in this PR, but we should consider either whether we can remove the need for this sorting entirely, or else look at the `order` frontmatter as well when sorting.

Fair point, I moved that part of the code.

I agree that the sorting shouldn't be necessary. I wrote a minimal version of nbdev (only library exporting) couple years ago. There I first iterated over all notebooks, collecting all the code cells into one datastructure that represented how the library should look. Then I had another logic turn that into files. This has some advantages, the major one being that it doesn't really matter in which order the notebooks are parsed. It would also make building things like the `__all__` collection easier. Would require a major re-write of the export mechanism though....

omlnaut:
@jph00 
Still new to this. The PR is still in state "Changes requested", even though I already pushed those. Do I need to do something else?

jph00:
Did you see this suggested way to implement it? I think this is neater, personally.
image

I think currently your .py and .ipynb files are out of sync. You should run `nbdev_export` to build the lib.

nbdev/doclinks.py
```diff
@@ -125,24 +123,24 @@ def nbglob_cli(
...
+# %% ../nbs/api/doclinks.ipynb 23
 @call_parse
 @delegates(nbglob_cli)
 def nbdev_export(
     path:str=None, # Path or filename
     **kwargs):
     "Export notebooks in `path` to Python modules"
     if os.environ.get('IN_TEST',0): return
-    files = nbglob(path=path, sort_by=lambda path_str: Path(path_str).name, **kwargs)
+    files = nbglob(path=path, **kwargs)#.sorted(lambda path_str: Path(path_str).name)
```
hamelsmu, line 134: Why is the call to sorted commented out here? Is that intentional?

omlnaut:
@jph00 I saw that, but I thought it was only meant to be a rough idea (not in correct syntax). Didn't know about that special function that the `L` has. I applied that in my PR.

@hamelsmu Nope, not sure what happened there. Is fixed by now.

Also, I pulled the master branch into mine because there was a release. I hope that was the right thing to do.

jph00:
Many thanks!

PR #1202, opened 2022-11-06
utility that creates a `requirements.txt` file from `settings.ini`

hamelsmu:
This is useful for things like creating HF spaces


nbdev/release.py
```diff
@@ -236,6 +236,15 @@ def write_conda_meta(path='conda'):
     _write_yaml(path, *_get_conda_meta())
 
 # %% ../nbs/api/release.ipynb 43
+def write_requirements(directory=None):
+    "Writes a `requirements.txt` file to `directory` based on settings.ini."
+    cfg = get_config()
+    d = cfg.config_path if not directory else Path(directory)
```
jph00, line 242: Extremely minor, but you don't need to `not` there

```suggestion
    d = Path(directory) if directory else cfg.config_path
```

nbdev/release.py
```diff
@@ -236,6 +236,15 @@ def write_conda_meta(path='conda'):
...
+    d.mkdir(exist_ok=True, parents=True)
```
jph00, line 243: If you use fastcore's `Path.mk_write` patch, you could avoid this line.

jph00:
Yay so happy to see this code @hamelsmu ! :D

hamelsmu:
@jph00 Thanks for the careful code review, appreciate it

PR #1210, opened 2022-11-08
fix: incorrect regex pattern for setting `output-file`

### LINKED ISSUE #1194: Notebook prefix is incorrectly removed when rendering URLs
I am building the `h3-landmarks` package using `nbdev` and while looking at the initial documentation generated, I see that the URL of the documentation strips numbers from it (`h3_landmarks.html` becomes `hlandmarks.html`), thus making the URL not very logical like we see in this example.

- Actual URL: https://bkowshik.github.io/h3-landmarks/hlandmarks.html
- Expected URL: https://bkowshik.github.io/h3-landmarks/h3_landmarks.html

See the same from the logs of `nbdev_preview` as well when I try it locally.
```
GET: /
GET: /hlandmarks.html
```

## References
- Medium page on why this library: https://bkowshik.medium.com/h3-landmarks-bce6cad971f9
- [WIP] Github repository of the library: https://github.com/bkowshik/h3-landmarks

seeM:
fixes #1194

### DIFF
--- a/nbdev/doclinks.py
+++ b/nbdev/doclinks.py
@@ -55,7 +55,7 @@ def _iter_py_cells(p):
         yield AttrDict(nb=nb, idx=int(idx), code=code, nb_path=nb_path, py_path=p.resolve())
 
 # %% ../nbs/api/doclinks.ipynb 11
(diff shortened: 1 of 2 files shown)


jph00:
@seeM afaict this is not the correct fix. The * seemed fine, for single digit prefixes. I think it just needs to be modified to only match the start of the name

PR #1215, opened 2022-11-08
add conda install doc to end-to-end tutorial

restlessronin:
I had some trouble with conda installation because the tutorial was incomplete. I added some instructions that would have helped me.


seeM:
Very helpful, thank you! :)

PR #1235, opened 2022-11-29
Make nbdev_create_config write the black_formatting setting

### LINKED ISSUE #1231: nbdev_create_config doesnt add black_formatting option
I am creating the config file via `nbdev_create_config --black_formatting True` but the option isn't added to the created _settings.ini_. I am a receiving a warning because I am using a private repo without having set the `GITHUB_TOKEN` but the same happens when I run the command in a completely empty repo.

On another note, it's not clear to me when the notebooks should be formatted? I hoped it would happen when saving the notebook but I couldnt trigger the black formatting after having added the setting manually?

dmose:
fixes #1231

### DIFF
--- a/nbdev/config.py
+++ b/nbdev/config.py
@@ -35,6 +35,9 @@ def _git_repo():
     except OSError: return
 
 # %% ../nbs/api/config.ipynb 11
+# When adding a named default to the list below, be sure that that name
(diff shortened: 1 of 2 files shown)


jph00:
Apologies I'm not sure how to fix this conflict, so I'll make the change at my end and credit you in the release notes - hope that's OK.

dmose:
No worries; thanks for landing it!

PR #1248, opened 2022-12-21
add extra info for users

slee5777:
Through debugging my nbdev project on deploying gh-page, the extra info will help others. 

Reference:  
https://forums.fast.ai/t/installing-nbdev-with-conda-on-windows-does-not-provide-executable-scripts/85692/8


environment.yml
```diff
@@ -0,0 +1,195 @@
+name: py_epi
```
seeM, line 1: Thanks for the PR Sarada! ❤️

Did you intend to include the `environment.yml` file? Would you mind removing it, then I'm very happy to merge this! 😄

slee5777: Sorry, my bad.  Now, the .yml file is deleted.

PR #1254, opened 2022-12-24
Fix `nbdev_update` when lib_name and lib_path are not the same

### LINKED ISSUE #1253: `nbdev_update` fails when `lib_name` is not the same as `lib_path`
The problem is here:

https://github.com/fastai/nbdev/blob/46d05d40fb8bfd1f6466f3daabc1fcf300a73fde/nbdev/sync.py#L33

It should be 

```python
 midx = import_module(f'{get_config().lib_path}._modidx') 
```

BirkhoffG:
This fixes #1253

### DIFF
--- a/nbdev/sync.py
+++ b/nbdev/sync.py
@@ -30,7 +30,7 @@ def absolute_import(name, fname, level):
 # %% ../nbs/api/sync.ipynb 7
 @functools.lru_cache(maxsize=None)
 def _mod_files():
-    midx = import_module(f'{get_config().lib_name}._modidx')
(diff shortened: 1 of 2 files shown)


BirkhoffG:
@jph00 @seeM @hamelsmu Are there any updates to this PR?

jph00:
Apologies I hadn't seen it - always a good idea to at-mention me when you pop a PR in so I get notified. I'll take a look now!

jph00:
There's a conflict that I can't resolve, so I'll make this change directly and manually credit you in the release notes - hope that's OK.

PR #1262, opened 2023-01-05
add 0 - Pre-planning status

ddobrinskiy:
the logic here is that setup.py uses indexing on the statuses list

since python is 0-indexed, putting `4` as status in `settings.ini` actually sets it at `5 - Production/Stable` instead of `4 - Beta` as one might expect, and so on


hugetim:
I can confirm this fix is needed. I had to set [my settings.ini](https://github.com/hugetim/nbstata/blob/master/settings.ini) to 3 to get it to show up as [4 on pypi](https://pypi.org/project/nbstata/).

jph00:
Many thanks!

PR #1265, opened 2023-01-09
Minor typo fix.

imflash217:
Changed `test all pass` to `passes all tests`


hugetim:
That's not a typo. The original is actually `tests all pass`, which is fine.

hamelsmu:
Yes not a typo

PR #1267, opened 2023-01-11
pipenv support: create pipfile from settings.ini

restlessronin:
I have been experimenting with pipenv and wanted a way to generate the pipfile from the settings.ini.

This is a first attempt at getting that done.

### DIFF
--- a/nbs/api/release.ipynb
+++ b/nbs/api/release.ipynb
@@ -301,20 +301,7 @@
    "cell_type": "code",
    "execution_count": null,
    "metadata": {},
-   "outputs": [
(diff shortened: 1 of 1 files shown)


jph00:
Thanks for the PR. I'd rather not add this -- there's a lot of different dependency systems out there and we're not going to support them all.

restlessronin:
sure. i get that. i might just create a helper lib for my own purposes and make it public.

restlessronin:
`nbdep` helper lib created. repo is at https://github.com/restlessronin/nbdep

PR #1268, opened 2023-01-12
add agplv3 license

joel-lbth:
could agplv3 be included, it's a good licence for software that might be used in cloud settings

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,6 +18,7 @@ licenses = {
     'mit': ('MIT License', 'OSI Approved :: MIT License'),
     'gpl2': ('GNU General Public License v2', 'OSI Approved :: GNU General Public License v2 (GPLv2)'),
     'gpl3': ('GNU General Public License v3', 'OSI Approved :: GNU General Public License v3 (GPLv3)'),
+    'agpl3': ('GNU Affero General Public License v3', 'OSI Approved :: GNU Affero General Public License (AGPLv3)'),
     'bsd3': ('BSD License', 'OSI Approved :: BSD License'),
 }
 statuses = [ '1 - Planning', '2 - Pre-Alpha', '3 - Alpha',


hamelsmu:
What's wrong with the Apache 2.0 License?

joel-lbth:
thanks hamel, yes, i think there's a relation between setup.py and settings.ini which twine uses (i think) to set the license on pypi.

jph00:
Many thanks.

joel-lbth:
thanks for this, i have pushed to pypi and all tickety-boo. nbdev has changed the way we work for the better, our team especially benefit from the improved versioning of notebooks which avoids merge conflicts caused by metadata changes, and the ease with which notebooks can be published as packages. i'll be promoting these benefits at fosdem 2023.

jph00:
Thanks for the nice feedback @joel-lbth . Where do you work? What are you using notebooks for in your team?

joel-lbth:
we work in local government and use notebooks to create visualisations for [council reports and publications](https://github.com/data-hamlets/census-data/tree/main/notebooks)

nbdev is useful for us because once we've got a few notebooks that want to share functions, and once there's a few people working on the same notebooks, modules, version control, docs, and tests all become important!

I used nbdev to make this a project called [data-mountains](https://github.com/joel-lbth/data-mountains) - it was a good experience, especially since it meant i could code entirely in notebooks, a very satisfying approach! working this way helped me to publish my first package on pypi.

jph00:
Splendid - thank you :)

rlgloden:
Hi.  Can someone explain why agplv3 was added?  agplv3 is makes software extremely hard to use in government or corporate world.  Our scanners just found the AGPL and flagged it as unusable.  Is anyone against backing this change out?

jph00:
nbdev is not agplv3.

rlgloden:
Thanks for looking into this. After reviewing the code again, I'm still unsure why 'agpl3' was added to the licenses dictionary. If nbdev isn't using the AGPL3 license, then the reference to 'agpl3' seems unnecessary. Moreover, its presence triggers false positives in government and corporate scanning tools that are designed to detect open source applications with restrictive licenses. Since this definition is not need for nbdev, it would be super helpful to remove the 'agpl3' reference to prevent unnecessary disputes with the IT department over the project's actual license when every a new version is released.

licenses = {
    'apache2': ('Apache Software License 2.0','OSI Approved :: Apache Software License'),
    'mit': ('MIT License', 'OSI Approved :: MIT License'),
    'gpl2': ('GNU General Public License v2', 'OSI Approved :: GNU General Public License v2 (GPLv2)'),
    'gpl3': ('GNU General Public License v3', 'OSI Approved :: GNU General Public License v3 (GPLv3)'),
    **'agpl3': ('GNU Affero General Public License v3', 'OSI Approved :: GNU Affero General Public License (AGPLv3)'),**
    'bsd3': ('BSD License', 'OSI Approved :: BSD License'),
}

jph00:
nbdev users can choose to license their software under agpl3 if they wish,
so it needs to be made available as an option to them. You're
misunderstanding this dictionary and it appears you are using a totally
broken tool or using it in the wrong way.

PR #1272, opened 2023-01-20
make enabling gh-pages mandatory

hamelsmu:
This is a docs only change @jph00 

See this discord conversation https://discord.com/channels/689892369998676007/887694559952400424/1065826828981391370

Looks like the GitHub api permissions changed here.  It was always a bit finicky, and I see lots of people getting stuck on this step -- I think it's better to make enabling pages more explicit.

### DIFF
--- a/nbs/tutorials/tutorial.ipynb
+++ b/nbs/tutorials/tutorial.ipynb
@@ -349,14 +349,23 @@
     "git add .\n",
     "git commit -m'Initial commit'\n",
     "git push\n",
-    "```"
+    "```\n",
+    "\n",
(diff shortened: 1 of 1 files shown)


jph00:
Many thanks @hamelsmu . Do we need to do anything to remove the auto-generation of gh pages as well?

hamelsmu:
> Many thanks @hamelsmu . Do we need to do anything to remove the auto-generation of gh pages as well?

https://github.com/fastai/workflows/commit/c0d2c7d3edbf7627ebd75312fbb4c6f4e37db22d

PR #1275, opened 2023-01-24
dirs_exist_ok has been moved inside _copytree

xl0:
As a follow-up to 898fba89b83387977c71b7fc4ee346ff0c4c8043


xl0:
@jph00 , would you like to take a look?

jph00:
Oops my bad - thanks for the fix!

PR #1280, opened 2023-02-01
Fix 'nbdev_conda' error when there are prior release tarballs

restlessronin:
There are errors caused by this assert

https://github.com/fastai/nbdev/blob/84c8be199f6ee8ca88118def64021425b7d69928/nbdev/release.py#L282

when previous versions of the distribution tarballs exist in the out directory. This fix deletes old tarballs before calling the conda build command.


jph00:
Thank you

PR #1283, opened 2023-02-03
add numbers back to files

hamelsmu:
This PR is in response to [this comment on the forums](https://forums.fast.ai/t/nbdev-v2-codebase-falls-short-of-literate-programming/103671/3).  I agree with the OP that is incredibly hard to read the source code without knowing what to read first.  Having the files sorted natively by the file system makes it a bit easier IMO.  The sorting of files also helps me.  

I commented on that forum post with a bit more context.  cc: @jph00


jph00:
Thanks @hamelsmu . It looks like you might need to run `nbdev_export` -- would you mind doing that?

Do we have an appropriate processor in place that will strip the numeric prefixes from the filenames so that they won't appear in URLs? (i.e so that existing external links to docs won't break)

hamelsmu:
Ohhh I didn't consider the URLs!!! However, I ran `nbdev_preview` locally and it looks like the prefix is stripped out of those URLs already 🎉

jph00:
Great thanks!

Issue #1284, opened 2023-02-04
Illegal Space `0xa0` Causes nbdev_export to Fail


JiaxiangBU:
I'm having an issue with the `nbdev_export` function failing due to an illegal space (`0xa0`) in
below.

To maximize the chances of my issue being fixed, I've created a minimal reproducible example in
contains the minimal amount of code needed to reproduce the error.

Here is the minimal reproducible example https://github.com/JiaxiangBU/nbdev-demo-230204

Here is the encoding error information using `nbdev_export`

```python
$ nbdev_export 
Traceback (most recent call last):
  File "d:\installed\miniconda\lib\runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "d:\installed\miniconda\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "D:\installed\miniconda\Scripts\nbdev_export.exe\__main__.py", line 7, in <module>
  File "d:\installed\miniconda\lib\site-packages\fastcore\script.py", line 119, in _f
    return tfunc(**merge(args, args_from_prog(func, xtra)))
  File "d:\installed\miniconda\lib\site-packages\nbdev\doclinks.py", line 137, in nbdev_export
    for f in files: nb_export(f)
  File "d:\installed\miniconda\lib\site-packages\nbdev\export.py", line 59, in nb_export
    mm.make(cells, all_cells, lib_path=lib_path)
  File "d:\installed\miniconda\lib\site-packages\nbdev\maker.py", line 210, in make
    write_cells(cells[last_future:], self.hdr, f)
  File "d:\installed\miniconda\lib\site-packages\nbdev\config.py", line 252, in write_cells
    if cell.source.strip(): file.write(f'\n\n{hdr} {cell.idx_+offset}\n{cell.source}')
UnicodeEncodeError: 'gbk' codec can't encode character '\xa0' in position 574: illegal multibyte sequence
```

deven367:
@JiaxiangBU I just tried it myself. It worked without a problem.

xl0:
@JiaxiangBU , please try my fix https://github.com/xl0/nbdev/tree/win-utf8

I don't have Windows, but pretty sure this is caused by utf-8 handling in Python on Windows.

`pip install git+https://github.com/xl0/nbdev@win-utf8`

deven367:
@xl0 @JiaxiangBU This is from the documentation. `nbdev` works on Windows but inside WSL.
[...]

Issue #1276, opened 2023-01-24
`nbdev_clean` fails to clean a notebook because it cannot decode some byte


hyunjongkimmath:
I have been having issues with `nbdev_clean`. I narrowed down the notebook that is having problems being cleaned and the issue seems to be with quotation characters. In the below Copyright message (based on that from the `nbdev` repo), the quotations around the word `License` seem to be the center of the problem:

```markdown
Copyright © 2019 onward Hyun Jong Kim. Licensed under the Apache License, Version 2.0 (the “License”); you may not use this project’s files except in compliance with the License. A copy of the License is provided in the LICENSE file in this repository.
```

The following is the error message:

```powershell
(trouver_py310_venv) C:\Users\hyunj\Documents\Development\Python\trouver>nbdev_clean --fname nbs/index.ipynb
C:\Users\hyunj\Documents\Development\Python\trouver_py310_venv\lib\site-packages\nbdev\clean.py:115: UserWarning: Failed to clean notebook
  warn(f'{warn_msg}')
C:\Users\hyunj\Documents\Development\Python\trouver_py310_venv\lib\site-packages\nbdev\clean.py:116: UserWarning: 'charmap' codec can't decode byte 0x9d in position 32929: character maps to <undefined>
  warn(e)
```

It may be relevant to note that I am using a Windows computer, the above error message was produced on command line, and that the editors I am using (VSCode to edit/read and Notepad++ to read) are set to encode files in UTF-8.

Beyond the block of text from the above diagnosis, the example that am working with is from [a library](https://github.com/hyunjongkimmath/trouver#readme) that I am developing using nbdev. The problem is not in the `master` branch (because I neglected to include a License statement), but rather is the `develop` branch.

hugetim:
Related:
https://forums.fast.ai/t/nbdev-export-raises-jsondecodeerror/103606/3

Myself, I was seeing the issue in the README.md generated by my index.ipynb. What triggered the error was trying to `pip install -e .`. Then the JSONDecodeError arose from this expression in setup.py: `long_description = open('README.md').read()`.

deven367:
@hyunjongkimmath I think the issue is arising from running nbdev on windows. Can you run it in WSL and see if the issue still persists?

PR #1288, opened 2023-02-06
pass encoding=utf8 when opening files

xl0:
I think this should solve #1284

https://dev.to/methane/python-use-utf-8-mode-on-windows-212i


hugetim:
May also solve #1276

jph00:
Do you mind resolving the conflicts? I think it's just a case of running the new nbdev_clean AFAICT.

xl0:
@jph00 , done.

Would you also like to merge https://github.com/fastai/nbdev-template/pull/9 ?

jph00:
> Would you also like to merge [fastai/nbdev-template#9](https://github.com/fastai/nbdev-template/pull/9) ?

Yes I would! :D (BTW, if you at-mention me on a PR then it goes to my inbox and won't be missed or forgotten.)

Issue #742, opened 2022-08-04
Allow a single notebook to export to multiple .py modules


jph00:
Use the same `#|export modname` syntax that was in nbdev1

hugetim:
The [directives docs](https://nbdev.fast.ai/explanations/directives.html#export-some.thing) suggest this issue can be closed, no?

jph00:
Maybe you could give it a try and see if it works? I don't recall where we got to with this.

hugetim:
It doesn't work in my simple example. The export directive silently fails, not exporting the function anywhere:
[nbdev-hello-world-test.zip](https://github.com/fastai/nbdev/files/10670462/nbdev-hello-world-test.zip)

turbotimon:
@hugetim it works in my example with nbdev Version: 2.3.13

*01_two.ipynb:*
```python
# | default_exp two
```

```python
#| export core
def foo_core_mixin(): # is exported to core.py
    pass
```

```python
#| export
def foo_two(): # is exported to two.py
    pass
```

However, I don't like the fact that it appears that you can't have two notebooks with the same `default_exp` (as it says here, it gets overwritten. However, i can't find anything in the docs about this)

PR #1291, opened 2023-02-07
Remove export-other from directives doc

hugetim:
See #742

### DIFF
--- a/nbs/explanations/directives.ipynb
+++ b/nbs/explanations/directives.ipynb
@@ -486,47 +486,6 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "### 📓 `#|export <some.thing>`"
-   ]
-  },
-  {
-   "cell_type": "markdown",
-   "metadata": {},
-   "source": [
-    "Similar to `#|export`, but instead of exporting to the module named by `#|default_exp` export to the specified module.\n",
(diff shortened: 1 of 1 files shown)


hugetim:
@jph00 , I made this quick pull request to make the docs consistent with the current functionality for now while we decide whether/how to implement the `#|export <other.module>` directive.

jph00:
Many thanks!

fgiasson:
Is there a place where this discussion is taking place? I faced that issue just now. It seems it was working with v1, but not anymore. Actually, it "kind" of work if:

 1. I create a new notebook with `default_exp foo`
 2. export one cell to default
 3. export one cell to `bar` 
 4. create `module/foo.py` manually
 5. create `module/bar.py` manually
 6. define `__all__` manually in `module/bar.py`
 7. run `nbdev_export` from the command line

Except for the manual process incurred, the exported cell(s) in `bar.py` are appended (instead of replaced) after each call of `nbdev_export` (but this duplication is not happening in `foo.py`, which uses the default export directive.

I haven't checked the code yet, but I would be interested to follow the conversation around that question from the maintenance team.

hugetim:
> Is there a place where this discussion is taking place? ...

This issue seems the proper place: https://github.com/fastai/nbdev/issues/742

PR #1292, opened 2023-02-08
nbdev_clean: Add trailing newlines to mask diff between Jupyter and VSCode

xl0:
Jupyter adds a trailing `\n` to images in cell outputs. Vscode-jupyter does not.
Notebooks should be brought to a common style by `nbdev_clean` to avoid unnecessary diffs.


jph00:
Looks good - thanks!

PR #1296, opened 2023-02-15
Don't reuse processes between notebooks

xl0:


jph00:
I think this will need an update to settings.ini to require the newest version of fastcore? (Which we'd need to release before this.)

xl0:
Yes

hamelsmu:
This PR seems stale, and is dependent on this PR which is also unfinished https://github.com/AnswerDotAI/fastcore/pull/515. 

I will close for now but please reopen if necessary.

PR #1301, opened 2023-02-20
Fix python version

BirkhoffG:
nbdev is supposed to support python>=3.7. It is redundant to keep updating `py_versions` every year with a new version of python.

We have already declared the minimal python version here:
https://github.com/fastai/nbdev/blob/923f2b010b5079642b84baa68688b745a4db40d8/setup.py#L50


BirkhoffG:
@jph00

jph00:
I believe the classifier section is for a different purpose, isn't it? Do you think it's not helpful to have the specific versions supported in that section?

PR #1312, opened 2023-03-13
allow trailing `:` in directives for YAML compliance

hamelsmu:
This is to allow people to work with Quarto a bit easier.  People can add a colon after their nbdev directives so it won't break Quarto.  This is the most minimal solution I can think of to unblock folks.   I tested this as follows:

- I added a unit test to make sure directives are still extracted correctly, and did some tests on nbdev
- I tested this on a stand alone quarto project and confirmed that adding a  trailing`:` alleviated Quarto's rendering issue.

This issue keeps cropping on the Quarto side.

This is a workaround for this issue: https://github.com/quarto-dev/quarto-cli/issues/3152

### DIFF
--- a/nbdev/process.py
+++ b/nbdev/process.py
@@ -33,6 +33,7 @@ def _quarto_re(lang=None): return re.compile(_dir_pre(lang) + r'\s*[\w|-]+\s*:')
 # %% ../nbs/api/03_process.ipynb 11
 def _directive(s, lang='python'):
     s = re.sub('^'+_dir_pre(lang), f"{langs[lang]}|", s)
+    if s.strip().endswith(':'): s = s.replace(':', '') # You can append colon at the end to be Quarto compliant.  Ex: #|hide:
     if ':' in s: s = s.replace(':', ': ')
     s = (s.strip()[2:]).strip().split()
     if not s: return None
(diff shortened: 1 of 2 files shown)


hamelsmu:
@cscheid just to confirm, if nbdev directives allowed a trailing `:` would that allow Quarto to do what it needs to do?  

IIUC 

`foo:` is equivalent to `foo: null` and `foo:` is valid YAML.  

cc: @jjallaire

jjallaire:
Note that while `#| export` by itself is valid YAML (the parser doesn't choke) it is interpreted as a string not an object so could cause other problems downstream.

`#| export:` by itself is valid but the value evaluates to `null` (which is a bit different than the intended usage which is `#| export: true`. I guess though if this directive is only used by nbdev then it's entirely fine for it to be interpreted on the quarto side as `null`.

hamelsmu:
This is a good point

> which is a bit different than the intended usage which is #| export: true

I suppose we might want to allow for that as well in a follow on PR, if folks want to abide by the semantic meaning of YAML.

Jeremy, WDYT?

hamelsmu:
@jph00 mind taking a look?  Thank you

jph00:
Thanks @hamelsmu , I think this is a reasonable workaround for now. Maybe in the future we should consider stuff like `#|export foo` to be the same as `#|export: foo`?

PR #1314, opened 2023-03-16
Fix encoding issue for Windows OS User especially for non-English user

JunDamin:
I am a windows os user in South Korea.
When I try to develop a small niche package using nbdev I faced an encoding problem.

As you may know, the default encoding for Korean in Windows OS is cp949. Therefore, when nbdev using Path.read_text from pathlib, their default encoding is cp949 and it causes a decoding error.

The simplest solution that I can figure out is explicitly set encoding for Path.read_text as 'utf-8'.

I believe it will not cause any change for most of users, at the same time it prevent not only Korean but also Japanese or other language users on Windows OS from encoding issues.

I hope this can help more Korean users enhance their productivity by nbdev.


jph00:
Thank you!

PR #1320, opened 2023-03-27
Doc TestPypi upload

itepifanio:
This PR documents how to use nbdev to upload the package to TestPypi and custom/private repositories


jph00:
Many thanks.

PR #1327, opened 2023-04-13
fixes #1232

### LINKED ISSUE #1232: `nbdev_conda` can generate `meta.yaml` that `conda-build` can not parse
To be clear, I'm pretty sure the fault is on `conda-build`: https://github.com/conda/conda-build/issues/4654

Should `nbdev` should still do something about it?
I don't see an obvious solution to the problem, would someone more knowledgeable in `pyyaml` be able to step in?

michaelaye:
Shuffles a few items around in the created meta.yaml to align it with what grayskull creates.
The resulting meta.yaml now (for me) builds without issues the conda package.

@jph00 please let me know if you'd rather like to have a complete grayskull invocation for creating the meta.yaml. Apparently you were open to that.


jph00:
Sorry I missed this! Yes I'd love to see a complete invocation :D

Issue #1323, opened 2023-03-29
The link to the documentation about nbdev_new with a private repo is broken (https://nbdev.fast.ai/cli.html#Using-nbdev_new-with-private-repos)


Taytay:
# Provide a minimally reproducible example

To repro, just run `nbdev_new` in a new repo that is private on Github.

You'll get this error:
```
/home/taytay/miniconda3/envs/nbdev/lib/python3.10/site-packages/ghapi/core.py:101: UserWarning: Neither GITHUB_TOKEN nor GITHUB_JWT_TOKEN found: running as unauthenticated
  else: warn('Neither GITHUB_TOKEN nor GITHUB_JWT_TOKEN found: running as unauthenticated')
Could not access repo: <name>/<repo_name> to find your default branch - `main` assumed.
Edit `settings.ini` if this is incorrect.
In the future, you can allow nbdev to see private repos by setting the environment variable GITHUB_TOKEN as described here:
https://nbdev.fast.ai/cli.html#Using-nbdev_new-with-private-repos
```
But this link is a 404: https://nbdev.fast.ai/cli.html#Using-nbdev_new-with-private-repos

It comes from:
https://github.com/fastai/nbdev/blob/693ad958add05b5c07d0998d90ad9d4e593fc9c6/nbdev/config.py#L90-L94

But I didn't know what link it was _supposed_ to go to.

nhstanley:
It looks like the instructions might have been moved to here(?): https://nbdev.fast.ai/api/release.html#setup

Either way, I would argue it is bad security practice to ever copy your token into a git repo directory, or keep multiple copies of a token around. There's just too much chance for error and leaking of keys[^1]. The user should be allowed to point nbdev to a key located elsewhere.

Alternatively, accessing it through [git-credential-manager](https://github.com/git-ecosystem/git-credential-manager) or some other credential manager would be preferred.

[^1]: And as a reminder, once something is in git, it's there forever unless you do some intentional surgery or restart with a fresh repo.

nhstanley:
After some more digging, it looks like nbdev tries to use [`ghapi`](https://github.com/fastai/ghapi), which expects the an environment variable `GITHUB_TOKEN` to be set.

It suggests putting you GitHub PAT in your bashrc/zshrc. 

```bash
export GITHUB_TOKEN=xxx
```

That's a solidly better solution, although I'm still not sure it's ideal.

PR #1331, opened 2023-04-22
Update GitHub PAT documentation

nhstanley:
Should address #1323.

It looks like the documentation is out of date, and recommends some insecure practices with the user's GitHub Personal Access Token. I've made some changes to the `nbs/api/01_config.ipynb` and `nbs/api/18_release.ipynb` notebooks that hopefully address these issues.


jph00:
Many thanks.

PR #1336, opened 2023-05-07
Introduces `#|expatch` directive

yegeniy:
Documentation is in `nbs/explanations/directives.ipynb`. Implementation is spread across `nbs/apis/0{1,4}*`.

**Description:**

This pull request proposes a new directive called `#|expatch`. It is similar to `#|export`, but it will also indent the content of the export cell by two spaces. This allows us to interleave multiple cells while defining certain methods.

**Use Case:**

This directive has been tested in production for a couple of months to develop [ETL](https://en.wikipedia.org/wiki/ETL) style [PySpark](https://spark.apache.org/docs/latest/api/python/) jobs.

Due to how the pyspark jobs are deployed, we end up needing all of the code wrapped in a method annotated with `@call_parse`. Using `#|expatch` has been helpful, because the code in the method is linear. It involves transforming complex, messy, and disparate sets of data. We interweave exported (`#|expatch`) code cells with documentation and example cells. But we also introduce documentation and examples along the way. There is plenty of code refactoring, helper methods, and code reuse. But at the end of the day, the bulk of the business logic is best kept inside a single exported method.

It has proven much easier and faster to debug and iterate this way. The alternative being that we declare the whole of the primary method inside a single hundred-line-plus cell and then test by running the whole thing or by reproducing parts of it in non-exported cells. With `#|expatch`, we can declare the exported code once and run it as we develop and test it.

**A Caveat on Usage:**

Please note that the `#|expatch` directive has very limited utility where it actually makes sense to refactor the code into small methods. It is not a good fit for those cases. But in that narrow set of use-cases where a long drawn-out method makes sense, this directive significantly simplifies documentation and debugging. In design pattern terminology, this is sometimes referred to as a "Transaction Script". Not to be confused with the much more common anti-pattern called "Long Method".


yegeniy:
There is also a [small thread in the nbdev discord about this idea](https://discord.com/channels/689892369998676007/1083213136070463558)

nbdev/export.py
```diff
@@ -24,9 +24,12 @@ def _exporti_(self, cell, exp_to=None): self.modules[ifnone(exp_to, '#')].append
     def _export_(self, cell, exp_to=None):
         self._exporti_(cell, exp_to)
         self.in_all[ifnone(exp_to, '#')].append(cell)
+    def _expatch_(self, cell, indent=2, exp_to=None):
+      cell.source = "\n".join([' ' * indent + line for line in cell.source.split("\n")])
+      self._export_(cell, exp_to)
```
yegeniy, line 29: This is the bulk of the change. Just to indent the code by 2 spaces.

nbdev/config.py
```diff
@@ -250,11 +250,12 @@ def add_init(path=None):
 
 # %% ../nbs/api/01_config.ipynb 51
 def write_cells(cells, hdr, file, offset=0):
-    "Write `cells` to `file` along with header `hdr` starting at index `offset` (mainly for nbdev internal use)."
+    "Write `cells` to `file` along with a space-indented header `hdr` starting at index `offset` (mainly for nbdev internal use)."
     for cell in cells:
-        if cell.source.strip(): file.write(f'\n\n{hdr} {cell.idx_+offset}\n{cell.source}')
+        indent = ' '*(len(cell.source)-len(cell.source.lstrip(" ")))
+        if cell.source.strip(): file.write(f'\n\n{indent}{hdr} {cell.idx_+offset}\n{cell.source}')
```
yegeniy, line 256: This fixes a small gotcha about how the nbdev internal comments are written into the generated file. These need to be indented by a couple of spaces as well.

yegeniy (commented):
@jph00 - I hope it's alright to nudge you about this PR. I left a couple of notes annotating the diff, in case it simplifies the review / decision.

hugetim:
Can you link to a sample notebook that uses this directive? It's not clear to me how it would work when running cells in the notebook. Does the code run cleanly when running notebook cells? (I'm puzzled how it could because code in a cell separated from the start of the method wouldn't have access to the variables created in the first method cell, within the method's namespace.)

If this directive were adopted, I'm thinking there would need to be an option to indent by four spaces rather than two, no?

ew-meetup:
> If this directive were adopted, I'm thinking there would need to be an option to indent by four spaces rather than two, no?

That would be a really nice improvement to the `expatch` directive!

There is already an `indent=2` argument to the `_expatch_` method, which defaults to 2 spaces. It would just be a matter of passing your desired indentation in to the directive processor. I recommend providing up to 3 ways of doing it:

- `#|expatch indent=4` at the directive level.
- `#|default_expatch_indent=4` at the notebook level.
- `default_expatch_indent=4` at the settings.ini level.

ew-meetup:
> Can you link to a sample notebook that uses this directive? It's not clear to me how it would work when running cells in the notebook.

I agree, it's a little nuanced. The 

In the discord chat thread, I included a couple of screenshots to demonstrate this:

The notebook
The exported code


> Does the code run cleanly when running notebook cells? (I'm puzzled how it could because code in a cell separated from the start of the method wouldn't have access to the variables created in the first method cell, within the method's namespace.)

Instead of writing your complete method in single cell, you would only include a stub in the `#|export`ed cell. Basically, the method signature and documentation. The rest of the method is `#|expatch`ed in.

yegeniy:
> If this directive were adopted, I'm thinking there would need to be an option to indent by four spaces rather than two, no?

That would be a really nice improvement to the `expatch` directive!

There is already an `indent=2` argument to the `_expatch_` method, which defaults to 2 spaces. It would just be a matter of passing your desired indentation in to the directive processor. I recommend providing up to 3 ways of doing it:

- `#|expatch indent=4` at the directive level.
- `#|default_expatch_indent=4` at the notebook level.
- `default_expatch_indent=4` at the settings.ini level.

yegeniy:
> Can you link to a sample notebook that uses this directive? It's not clear to me how it would work when running cells in the notebook.

I agree, it's a little nuanced. I attached an example in this comment. But first, please note the "Caveat on Usage":

> Please note that the #|expatch directive has very limited utility where it actually makes sense to refactor the code into small methods. It is not a good fit for those cases. But in that narrow set of use-cases where a long drawn-out method makes sense, this directive significantly simplifies documentation and debugging. In design pattern terminology, this is sometimes referred to as a "Transaction Script". Not to be confused with the much more common anti-pattern called "Long Method".

In the discord chat thread, I included a couple of screenshots to demonstrate this:

The notebook

⚠️ take particular note of the second cell, which is not exported. It is just there to set local testing values for the arguments.

The exported code

yegeniy:
> Does the code run cleanly when running notebook cells? (I'm puzzled how it could because code in a cell separated from the start of the method wouldn't have access to the variables created in the first method cell, within the method's namespace.)

Yes, it executes cleanly, but only if is a specific pattern is followed first. Mainly, the arguments to the method need to be declared in a separate cell once the method is declared. This is useful for demonstrating your code.

Instead of writing your complete method in single cell, you would only include a stub in the `#|export`ed cell. Basically, the method signature and documentation. The rest of the method is `#|expatch`ed in.

hugetim:
> ...narrow set of use-cases...

Thank you. Your responses help me understand better how it works for you in practice. My own opinion (just the opinion of one other nbdev user), having considered it a bit more, is that this would be better as an [nbdev plugin](https://nbdev.fast.ai/tutorials/extensions.html) rather than as part of nbdev proper. The reason is that it introduces 1-2 new directives and/or a new setting for the sake of a fairly narrow set of use cases. 

> ...limited utility where it actually makes sense to refactor the code into small methods. It is not a good fit for those cases.

There's also a risk of people using this directive in situations for which it is not a good fit rather than refactoring their code into smaller functions, inadvertently encouraging worse code. Providing it instead as a plugin reduces this risk, I think.

jph00:
I apologise for sitting on this for so long - I've had trouble getting my head around what it's for, and this discussion is very useful thank you! I agree that it would be best made available as a plugin in a separate project, so I'll close this PR.

yegeniy:
Brilliant. Thanks for the suggestion @hugetim. I am not sure how I missed
the plugin support in nbdev, but I am grateful for the link.

I believe this _should be_ possible using processors. But it may not be,
we’ll see. On the surface it seems like a better fit for any custom
directive, for sure though.

PR #1350, opened 2023-06-07
add repo and tag options to nbdev_new`

dsm-72:
See template [issue 13](https://github.com/fastai/nbdev-template/issues/13)


dsm-72:
@seeM does this work as I think it does?

jph00:
Many thanks! Sorry for the slow merge.

PR #1351, opened 2023-06-07
added conda_user and package_data to cfg_tail

dsm-72:
As stated in the title. Since a main features of `nbdev` is packaging it makes sense to at least include `conda_user` (only knew to add it after running into a complaint from `nbdev_conda`).  I also have a [PR](https://github.com/fastai/nbdev-template/pull/15) on `fastai/nbdev-template`  to handle package data as it seems that it was otherwise ignored.

### DIFF
--- a/nbs/api/01_config.ipynb
+++ b/nbs/api/01_config.ipynb
@@ -1,6 +1,7 @@
 {
  "cells": [
   {
+   "attachments": {},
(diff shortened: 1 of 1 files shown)


dsm-72:
@seeM anything else I should do?

jph00:
Good idea - thanks.

PR #1356, opened 2023-07-03
update install_quarto to distinguish arm64 and amd64

JonasHarnau:
_install_linux now infers the machine and installs the appropriate quarto version accordingly.

Addresses issue for example when running a devcontainer for development with VSCode.

Concrete case using the image `mcr.microsoft.com/devcontainers/python:1-3.11-bullseye` on an Apple M2 machine. This caused the quarto installation to fail since `sys.platform` in the devcontainer is linux but the arm64, not amd64, distribution of Quarto is needed.


jph00:
Many thanks!

Issue #1382, opened 2023-11-07
ImportError: cannot import name 'uname' from 'os' [Windows]


andruum:
In new version 2.3.13 there is a bug of os lib import (for example, when command nbdev_new is used), I guess this was not tested in Windows environment:

nbdev_new 

site-packages\nbdev\quarto.py", line 6, in <module>    from os import system, uname
ImportError: cannot import name 'uname' from 'os'

Version 2.3.12 works fine.

deven367:
@andruum Since, `nbdev` is not intended to be used on Windows, you could try to initialize a new project through, WSL.
![image](https://github.com/fastai/nbdev/assets/31466137/0fa02336-1497-4fe7-a761-b70fcdf4b26c)

sei-msd:
nbdev_export will export CRLF anyway.

andruum:
> nbdev_export will export CRLF anyway.

But how is it related to the issue? 

I proposed a fix in the pull request, it works fine for my team at least, it works.

PR #1385, opened 2023-11-15
moved uname into func to fix Windows import error

andruum:
Fixed problem for Windows, which I described below, by simply hiding import inside necessary function.

https://github.com/fastai/nbdev/issues/1382


jph00:
Thank you!

Issue #1389, opened 2023-12-11
`proc_nbs` does not make use of its arguments `file_glob` and `file_re`


lukastk:
Currently the arguments `file_glob` and `file_re` are not being used in `nbdev.serve.proc_nbs`. See the code below

```python
def proc_nbs(
    path:str='', # Path to notebooks
    n_workers:int=defaults.cpus,  # Number of workers
    force:bool=False,  # Ignore cache and build all
    file_glob:str='', # Only include files matching glob
    file_re:str='', # Only include files matching glob
    **kwargs):

    # ... code ...

    files = nbglob(path, func=Path, file_glob='', file_re='', **kwargs)
    if (path/'_quarto.yml').exists(): files.append(path/'_quarto.yml')
    if (path/'_extensions').exists(): files.extend(nbglob(path/'_extensions', func=Path, file_glob='', file_re='', skip_file_re='^[.]'))

    # ... code ...
```

PR #1390, opened 2023-12-11
Changed `serve.proc_nbs` so that it makes use of its arguments `file_glob` and `file_re`

lukastk:
Fixed the issue below

[#1389](https://github.com/fastai/nbdev/issues/1389)


jph00:
Many thanks :D

jph00:
I had to revert this because it broke website rendering.

Issue #1393, opened 2023-12-29
Nested `lib_path` values break module -> notebook sync


scwe:
When running `nbdev_update` on a project that uses a nested `lib_path` setting (i.e. `path/to/file`) the update will fail

Minimal reproduceable example is available [here](https://github.com/scwe/nbdev-update-bug/tree/main). The exact error I get here is
```
Traceback (most recent call last):
  File "/home/scott/.local/bin/nbdev_update", line 8, in <module>
    sys.exit(nbdev_update())
  File "/home/scott/.local/lib/python3.10/site-packages/fastcore/script.py", line 119, in _f
    return tfunc(**merge(args, args_from_prog(func, xtra)))
  File "/home/scott/.local/lib/python3.10/site-packages/nbdev/sync.py", line 79, in nbdev_update
    files = files.filter(lambda x: str(Path(x).absolute().relative_to(lib_dir) in _mod_files()))
  File "/home/scott/.local/lib/python3.10/site-packages/fastcore/foundation.py", line 162, in filter
    return self._new(filter_ex(self, f=f, negate=negate, gen=False, **kwargs))
  File "/home/scott/.local/lib/python3.10/site-packages/fastcore/basics.py", line 642, in filter_ex
    return list(res)
  File "/home/scott/.local/lib/python3.10/site-packages/nbdev/sync.py", line 79, in <lambda>
    files = files.filter(lambda x: str(Path(x).absolute().relative_to(lib_dir) in _mod_files()))
  File "/home/scott/.local/lib/python3.10/site-packages/nbdev/sync.py", line 38, in _mod_files
    midx = import_module(f'{get_config().lib_path.name}._modidx')
  File "/usr/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 992, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1004, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'nested'
```

PR #1394, opened 2023-12-29
Fix sync for nested libraries

scwe:
A fix for https://github.com/fastai/nbdev/issues/1393. In essence this fixes it by using the complete library path for importing the modidx file rather than just grabbing the lowest level subdirectory from the path.

I'm not entirely sure this is the _right_ fix for the problem, but in the event that it is hopefully it makes your lives very easy.

### DIFF
--- a/nbdev/sync.py
+++ b/nbdev/sync.py
@@ -19,7 +19,7 @@ from fastcore.script import *
 from fastcore.xtras import *
 
 import ast
-from importlib import import_module
+import importlib
 
 # %% ../nbs/api/06_sync.ipynb
 def absolute_import(name, fname, level):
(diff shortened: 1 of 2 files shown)


scwe:
Hey @jph00 and/or @hamelsmu :wave: I'm not sure who best to ping on this to get some eyes on this so we can start using `nbdev_update` in our project.

jph00:
Sorry I missed this old PR! Have you had much of a chance to test it in the meantime? If you're pretty confident it works, I'll merge it - I don't use the sync feature myself so I'm not confident I know what to look for!

scwe:
> Have you had much of a chance to test it in the meantime?

Wow, what a coincidence - we actually _just_ decided internally to switch to the fork which contains this commit. So I haven't tested it much except for in the example repo. 

How would you feel about leaving this open and giving us ~1 month to hammer on it a little internally and then I'll ping you assuming all goes well?

I think that'll give me more confidence that everything is working as expected.

jph00:
Sure - take your time :)

scwe:
@jph00 I think I'm comfortable with saying this is working well. We've been using it as part of CI for a few weeks and I've also retested to make sure I didn't break flat structured projects in the process.

jph00:
Many thanks for the update!

PR #1397, opened 2024-02-19
Fixes for Python 3.12 pkg_resources

velaia:
Fixing ModuleNotFoundError with Python 3.12.2:

`ModuleNotFoundError: No module named 'pkg_resources'`

Similar to https://github.com/fastai/fastcore/pull/542 (deprecation of pkg_resources)


jph00:
I'll hold off on this until we can find a fix for the problem in `fastcore`.

PR #1400, opened 2024-03-22
Add --no_browser option to nbdev_preview

bamford:
This is useful to avoid trying to open a browser (and associated error messages) when working on a remote machine, e.g. via ssh tunnels for Jupyter and Quarto.

Quarto already has this option, so it is simply a case of passing it along.


jph00:
Many thanks!

PR #1402, opened 2024-03-29
Added function update settings.ini from existing requirements.txt file

AndreasThinks:
As I often have pre-existing requirements.txt files, manually updating my settings.ini file can be a bit of a pain.  I've written a bit of a helper function for this (essentially the opposite of release.write_requirements by @hamelsmu)

Not sure if useful for others, so as draft for discussion?


jph00:
There's already a `write_requirements` -- probably better to use that direction instead.

PR #1419, opened 2024-06-13
Fix raw strings containing escape sequences

### LINKED ISSUE #1396: showdoc: SyntaxWarning: invalid escape sequence with Python 3.12
# Provide a minimally reproducible example

run nbs/api/08_showdoc.ipynb with Python 3.12 kernel

# some background

what I understand from https://stackoverflow.com/questions/77531208/python3-12-syntaxwarning-on-triplequoted-string-d-must-be-d is that with python 3.12, these messages have been promoted from DeprecationWarning to SyntaxWarning and are therefore not automatically filtered anymore.

image

This is not a bug but I guess something should be made about it?

imagejan:
Closes #1396.


jph00:
Thanks! Can you please modify the source notebooks and export them?

jph00:
Thank you!

PR #1432, opened 2024-08-14
fix: request jupyterlab installation, not jupyter (outdated metapackage)

corneliusroemer:
The fact that the error message said `jupyter` not `jupyterlab` made me install the wrong package. `jupyter` is a really outdated metapackage (last update 2015).


jph00:
I don't want to recommend lab - I for one don't use it, and I know a lot of our community doesn't either. I only intended 'jupyter' here to mean 'whatever version of jupyter you like'!

PR #1436, opened 2024-08-22
Feature: add support for .py files in py:percent format

lukastk:
- Introduces the ability to handle .py files as py:percent files within nbdev.
   - Does this by converting .py files to .ipynb format as an intermediate step in `NBProcessor.__init__`
- Added example of usage in 'nbs/api/04_export.ipynb'.

Can export all *.py files by running `nbdev_export --file_glob '*.py'`

See https://jupytext.readthedocs.io/en/latest/formats-scripts.html for info on the `py:percent` format.


lukastk:
Added issue here:
https://github.com/fastai/nbdev/issues/1437

jph00:
This is a great feature! But I don't want to add these additional deps to nbdev. Could we make it an extension or some other kind of separate project? Or alternatively only enable these new features if these deps are installed (but not make them requirements)?

PR #1448, opened 2024-09-17
test if in nbdev_repo & improve `nbdev_export` err message

hamelsmu:
@jph00 IIUC you suggested making this function public

```python
def _nbdev_config_file(cfg_name=_nbdev_cfg_name, path=None):
    cfg_path = path = Path.cwd() if path is None else Path(path)
    while cfg_path != cfg_path.parent and not (cfg_path/cfg_name).exists(): cfg_path = cfg_path.parent
    if not (cfg_path/cfg_name).exists(): cfg_path = path
    return cfg_path/cfg_name
```

But you still have to check if that path exists.  The way to check if nbdev exists might not be as apparent to users so maybe a simple function that helps you check (which is named appropriately) is better, given its usefulness?

Also, I improved the error message with `nbdev_export`


hamelsmu:
Ahh maybe you just want to use `nbdev_config_file().exists()` directly instead?

jph00:
Looks like CI problems @hamelsmu ! :D

jph00:
@hamelsmu did you want to mark this draft, or is it ready?

nbdev/doclinks.py
```diff
@@ -143,9 +143,10 @@ def nbdev_export(
     **kwargs):
     "Export notebooks in `path` to Python modules"
     if os.environ.get('IN_TEST',0): return
+    if not is_nbdev(): raise Exception('`nbdev_export` must be called from a directory within a nbdev project.')
     if procs:
-      import nbdev.export
-      procs = [getattr(nbdev.export, p) for p in L(procs)]
+        import nbdev.export
```
hamelsmu, line 148: My jupyter notebook got angry with the indentation here.  So I fixed it FYI

hamelsmu:
@jph00 it is ready now 😄

PR #1454, opened 2024-09-24
Watch export

Isaac-Flath:
@jph00 

Created `watch_export` to run `nb_export` on any modified notebook in the directory being watched. 
 Uses nbdev config if available by default.


jph00:
Very nice! :D

PR #1456, opened 2024-10-01
Export on modified only bug fix

Isaac-Flath:
@jph00 

This is the bug that you helped us debug—changed `watch_export` to only export on file modifications to prevent infinite loop on linux.


jph00:
Great :)

PR #1460, opened 2024-10-19
Html show doc rendering

Isaac-Flath:
Improved `show_doc` rendering for HTML, including rending the documents table as an HTML table.  This is useful in order to use `show_doc` with FastHTML apps.

I cloned every fastai repo and checked for uses of the BasicHTMLRenderer that is being changed, and there were no uses (see result in image).

Screenshot 2024-10-18 at 11 24 25 PM


Isaac-Flath:
@jph00 - I am having trouble figuring out what to do about the failed test.  It seems to just effect ubuntu python 3.9 and not other version/OS combos.  It seems like the tests passed for the nbdev library, but Quarto is running into an error when trying to render the docs in the `test docs build` on the `tutorial.ipynb` notebook.

When I tried to research the error it looked like similar errors could be caused by incorrect quarto directives around multi-column layouts, however I don't think that's the case here since I can't find any multi-column directives that would cause this.

nbdev/showdoc.py
```diff
@@ -192,24 +192,49 @@ def show_doc(sym,  # Symbol to document
...
+        sig = _fmt_sig(self.sig) if self.sig else ''
+        # Escape < and > characters in the signature
+        sig = sig.replace('<', '&lt;').replace('>', '&gt;')
+        doc += f'<blockquote><pre><code>{self.nm} {sig}</code></pre></blockquote>'
+        if self.docs:
+            doc += f"<p><i>{self.docs}</i></p>"
+        if self.dm.has_docment:
+            doc += _create_html_table(str(self.dm))
         return doc
+    __repr__=__str__=_repr_html_
```
jph00, line 232:
```suggestion
```

Isaac-Flath:
Closing for debugging, will reopen when resolved

Isaac-Flath:
Closing in order to create a clean PR with changes

PR #1463, opened 2024-10-20
Html renderer improvements

Isaac-Flath:
New clean PR for HTML rendering after the CI madness.  

I also changed it to address the source link removal issue.


jph00:
Thank you!

PR #1469, opened 2024-11-19
Improve backtick regex matching for function names

ncoop57:
This PR improves the regex detection for doclinks in nbdev. Specifically it now handles parenthesis being included in backticks:

```python
md = """This is a link to `numpy.array` and to `get_config()` but not a link to `foobar`.
And not a link to <code>dict2nb</code>.

    This is not a link to `get_config`
"""
print(NbdevLookup('nbdev').linkify(md))
```
```markdown
This is a link to [`numpy.array`](https://numpy.org/doc/stable/reference/generated/numpy.array.html#numpy.array) and to [`get_config()`](https://nbdev.fast.ai/api/config.html#get_config) but not a link to `foobar`.
And not a link to <code>dict2nb</code>.

    This is not a link to `get_config`
```

Additional tests were also added to check against edge cases.

### DIFF
--- a/nbdev/doclinks.py
+++ b/nbdev/doclinks.py
@@ -236,12 +236,13 @@ class NbdevLookup:
         _,py,gh = res
         line = _lineno(sym, py)
         return f'{gh}#L{line}'
-
+    
     def _link_sym(self, m):
         l = m.group(1)
         s = self.doc(l)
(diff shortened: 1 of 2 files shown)


ncoop57:
@jph00

jph00:
Great stuff @ncoop57 :D

PR #1471, opened 2024-11-20
Update NbdevLookup to support import aliases and improve docstrings

ncoop57:
This PR adds support to us NbdevLookup on import aliases such as `import numpy as np`. For example, you can now do the following:

```python
import numpy as np
NbdevLookup(ns=globals()).linkify('this is an aliased import link `np.array`')
```
```markdown
'this is an aliased import link [`np.array`](https://numpy.org/doc/stable/reference/generated/numpy.array.html#numpy.array)'
```

The way this works, is that an alias mapping is created in NbdevLookup:
```python
self.aliases = {n:o.__name__ for n,o in (ns or {}).items() if isinstance(o, ModuleType)}
```

This maps the alias to its true name. We then use this in the get item function when necessary:
```python
def __getitem__(self, s): 
        if '.' in s:
            pre,post = s.split('.', 1)
            if pre in self.aliases: s = f"{self.aliases[pre]}.{post}"
        return self.syms.get(s, None)
```

However, it required removing `lru_cache(None)` as a class level decorator since the dictionary for the namespace used to create the aliases of the imports and their true symbol names is not hashable. A work around would be to not do it on the class level and just decorate the functions that need it, but wanted to open up the PR to start a discussion on this as a direction before investing more time into it.


ncoop57:
@jph00

ncoop57:
I added a commit that separates out the building of the lookup table from the init function of nbdev lookup class, as discussed with @jph00. Also added some test cases that evaluate whether or not the caching is actually working. I think it's ready for primetime!

jph00:
Super great @ncoop57 ! :D

PR #1478, opened 2024-12-02
Migrate index creation from nbdev-index

ncoop57:
This PR adds the function `create_index`, which has been migrated from the `nbdev-index` repo. Additionally, a change was made to this index creation to allow for built-ins to be properly indexed from python's standard lib. However, in order to use this function Sphinx needs to be installed (`pip install sphinx`). It has been made an optional dependency, but has been added to the dev requirements in order for CI/CD to work. This PR is a dependency for the [PR](https://github.com/AnswerDotAI/nbdev-index/pull/39) that I have in `nbdev-index`. So, I think the order will matter here - you'll need to merge this one first, and then `nbdev-index` will need to be merged.


ncoop57:
@jph00 not sure why this one CI test is running for so long. Any ideas?

jph00:
> @jph00 not sure why this one CI test is running for so long. Any ideas?

Just manually restarted it.

jph00:
Weird. Seems to be stuck in the same place. I'll merge this and check locally. BTW I think the nb changes you made won't look great in the docs -- maybe you can clean up that a bit so this new functionality has clearer documentation and easier to understand tests?

PR #1482, opened 2024-12-28
Change symbols dict logic to prioritize builtins

ncoop57:
This PR updates the logic for the creation of the lookup table for the python symbols in the nbdev docs api for linkifying markdown to prioritize builtins over other standard library modules.


jph00:
Thank you! :)

PR #1485, opened 2025-01-16
add llms.txt

radekosmulski:


nbs/llms.txt
```diff
@@ -0,0 +1,50 @@
...
+
+## Tutorials
+
+- [Getting started](https://nbdev.fast.ai/getting_started.html.md)
+- [Tutorial](https://nbdev.fast.ai/tutorials/tutorial.html.md)
+- [Best practices](https://nbdev.fast.ai/tutorials/best_practices.html.md)
+
+## API
+
+- [API List](https://nbdev.fast.ai//apilist.txt): A succint list of all functions and methods in nbdev.
```
hamelsmu, line 22: I believe this url is incorrect?  Has an extra slash in the path?

hamelsmu (changes requested):

nbs/llms.txt
radekosmulski: fixed

jph00:
Awesome @radekosmulski ! :D

Issue #1480, opened 2024-12-19
`nbdev_clean` documentation partly wrong


ritagonmar:
I was trying out  `nbdev_clean` and I realized the documentation is not completely right, which caused me some confusion.
In the documentation (https://nbdev.fast.ai/api/clean.html#nbdev_clean) it says:
>fname 	str 	None 	A notebook name or glob to clean

but this is not true, it accepts a single notebook name or a path, but not a glob (see detailed below). Maybe this should be changed to avoid confusion, cause it took me a while to understand what works and what doesn't, and I only figured with the help of old issues.

what works:
- one single file (e.g. `nbdev_clean --fname nb1.ipynb`)
- a path (e.g. `nbdev_clean --fname ./`)

what does not work:
- passing several files (e.g. `nbdev_clean --fname nb1.ipynb nb2.ipynb`)
- a glob (e.g. `nbdev_clean --fname *.ipynb`)
- a path with a glob (e.g. `nbdev_clean --fname ./*.ipynb`)

Very cool library and features otherwise!

PR #1488, opened 2025-01-25
Allow `nbdev_clean` to accept multiple filenames or globs (#1480)

jbwhit:
## Overview

This PR addresses [#1480](https://github.com/AnswerDotAI/nbdev/issues/1480) by allowing `nbdev_clean` to accept **multiple filename arguments**, rather than just a single file or glob. This makes it easier to clean multiple notebooks in one command.

## What Changed

- **Multiple Files Support**: Updated `nbdev_clean` so it can handle either a single path/glob string or a list of paths/globs.
- **New Tests**: Added tests confirming multiple files are handled correctly (they fail on the old code, pass on the new).
- **Doc Update**: Revised the docstring to reflect this new usage.

## Known Similar Issue in `nbdev_test`

`nbdev_test` (and possibly other CLI commands) still only accept single-file arguments. This PR focuses on `nbdev_clean` specifically to keep it scoped. Let me know if you want a follow-up PR for `nbdev_test`.

## Skipped a Failing Test in `10_processors.ipynb`

- One cell running `_run_procs([add_show_docs, exec_show_docs])` currently fails for unrelated reasons.
- I added `#|hide` and `#|eval: false` to skip that cell so `nbdev_prepare` can complete.
- We can address that failing test in a separate issue or PR.

**Please let me know if you have any questions or suggestions—this is my first PR for this project!**

### DIFF
--- a/nbdev/clean.py
+++ b/nbdev/clean.py
@@ -130,7 +130,7 @@ def _nbdev_clean(nb, path=None, clear_all=None):
 # %% ../nbs/api/11_clean.ipynb
 @call_parse
 def nbdev_clean(
-    fname:str=None, # A notebook name or glob to clean
+    fname=None, # A notebook path or a list of paths/globs to clean
     clear_all:bool=False, # Remove all cell metadata and cell outputs?
     disp:bool=False,  # Print the cleaned outputs
     stdin:bool=False # Read notebook from input stream
@@ -141,7 +141,12 @@ def nbdev_clean(
     _write = partial(process_write, warn_msg='Failed to clean notebook', proc_nb=_clean)
     if stdin: return _write(f_in=sys.stdin, f_out=sys.stdout)
     if fname is None: fname = get_config().nbs_path
-    for f in globtastic(fname, file_glob='*.ipynb', skip_folder_re='^[_.]'): _write(f_in=f, disp=disp)
+    # If `fname` is a single string, wrap it in a list.
+    # Otherwise assume it's already a list/tuple of paths/globs.
+    if isinstance(fname, (str, Path)): fname = [str(fname)]
+    for f in fname:
+        for nb_path in globtastic(f, file_glob='*.ipynb', skip_folder_re='^[_.]'):
+            _write(f_in=nb_path, disp=disp)
 
 # %% ../nbs/api/11_clean.ipynb
 def clean_jupyter(path, model, **kwargs):
--- a/nbs/api/10_processors.ipynb
+++ b/nbs/api/10_processors.ipynb
@@ -739,6 +739,8 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "#|hide\n",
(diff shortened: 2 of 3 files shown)


hamelsmu:
> currently fails for unrelated reasons.

Please tell us more.  What are the reasons?

jbwhit:
Thank you for asking about this issue. I was able to resolve the failing test by updating my pandas installation with `pip install --upgrade pandas`.

The test in 10_processors.ipynb with _run_procs([add_show_docs, exec_show_docs]) is now passing, and I've removed the `#|hide` and `#|eval: false` markers. It seems it was just an environment-specific issue on my end rather than a problem with the codebase.

I'll update the PR shortly to remove those markers and ensure all tests are running properly.

jbwhit:
Ok I've updated the PR -- all tests now pass.

jph00 (changes requested):
Thank you for this PR. IIUC, `_run("git config user.email 'nbdev@fast.ai'")` will actually modify the git config of a user running the tests. Tests should not modify a user's environment. Also BTW you might find fastcore's `listify` helpful.

jbwhit:
I can make a different test without using `_run("git config user.email 'nbdev@fast.ai'")` -- the reason I included it was that was trying to base my new test off existing tests, and this code already exists in the same notebook (`nbs/api/11_clean.ipynb`). If this code shouldn't be there at all I can make a new PR that attempts to write the test correctly?

I will pare down my test to the minimum and I will look to use `listify` as well. Thanks for your attention and for creating this project in the first place!

jph00:
Oh wow I'd completely missed that! Thanks for noticing. Yes we should get
rid of it :D

jbwhit:
I've updated the PR to enhance `nbdev_clean` to properly handle multiple notebook paths.

1. **The issue**(https://github.com/AnswerDotAI/nbdev/issues/1480): When passing a list of paths to `nbdev_clean`, it fails because `globtastic` expects a string or Path object, not a list. This PR allows nbdev_clean to accept multiple filename arguments, rather than just a single file or glob.

2. **The fix**: Use `fastcore.listify` to properly handle both single paths and lists of paths, ensuring that `globtastic` receives the correct type. I changed the code back to something much closer to the original implementation that my initial commits.

3. **Added test**: I've added a comprehensive validation test that:
   - Demonstrates that the current implementation fails when given a list of paths
   - Shows that the fixed implementation using `listify` correctly processes all paths
   - Verifies that proper cleaning occurs on all specified notebooks
   - Does not require the usage of git within the test at all

The test shows that without `listify`, the function fails when given a list of paths, but works correctly with the fix. This change maintains backward compatibility while adding the ability to pass multiple paths.

I've kept this PR focused solely on fixing the `nbdev_clean` functionality and will address the git test refactoring in a separate PR as suggested.

jph00:
Many thanks!

PR #1489, opened 2025-01-30
Add exporting single notebook outside nbdev project

radekosmulski:
To export a notebook to a .py file outside an nbdev project:
```
import nbdev
nbdev.export.nb_export('./example.ipynb', '.', solo_nb=True)
```

Key change: added `solo_nb` arg that propagates the intent through functionality involved in exported module generation.

Example:
image


jph00:
Thank you! :D

PR #1490, opened 2025-02-05
Better handling of cells that contain string 'nbdev_export('

ozpau:
Currently, if you just happen to include string `nbdev_export(` in your code cell, the whole cell will be silently ignored by `nbdev_prepare`, `nbdev_test`, `nbdev_proc_nbs` and `nbdev_preview`. If this cell happens to contain code that is exported or rendered in docs, it would result in the corresponding command failing with hard to decipher error.

I have removed the corresponding check from the nbdev code. I am not sure we need it, but I can make a version of this pull request that handles that more gracefully by injecting monkey patching code into notebooks, something like this:
```python
k.run_cell('import nbdev\ndef noop(*args, **kw): pass\nnbdev.nbdev_export = noop')
```

I imagine the check was there to prevent infinite loops of `nbdev_prepare` calling itself, but this doesn't happen because the check is for `nbdev_export` and not for `nbdev_prepare`.

### DIFF
--- a/nbdev/processors.py
+++ b/nbdev/processors.py
@@ -221,7 +221,7 @@ def _ast_contains(trees, types):
 
 def _do_eval(cell):
     if cell_lang(cell) != 'python': return
-    if not cell.source or 'nbdev_export'+'()' in cell.source: return
+    if not cell.source: return
     trees = cell.parsed_()
     if cell.cell_type != 'code' or not trees: return
     if cell.directives_.get('eval:', [''])[0].lower() == 'false': return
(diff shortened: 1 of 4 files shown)


hamelsmu:
Can you give me an example use case where you need to have the string `nbdev_export()` in a code cell ?  Is this something that is blocking you or more theoretical?

ozpau:
I have stumbled upon this while creating a tool to generate notebook from a template. https://github.com/ozpau/oztools/blob/main/nbs/api/02_nbd.ipynb see `new_notebook_template`.

It's not a blocker, since there's a simple solution to instead write something like `'nbdev.nbdev_e'+'xport()'` in the code cell (and that's what nbdev uses internally).

But it was confusing at first because my notebook run, but exported code was wrong.

PR #1491, opened 2025-02-06
Fix llmctxissue

hamelsmu:
IIUC `settings.ini` can be ignored by `_proc_file` which will result it in it not getting copied to the destination directory, which is what we want to avoid because they end up in `_proc` which confuses llms-txt cli, when notebooks are in the root of the repo.


jph00:
Wow nice find!

PR #1494, opened 2025-02-06
properly escape json and yaml on `nbdev_new`

### LINKED ISSUE #1493: When generating index.ipynb, cell content is not sanitized
# Steps to reproduce
1. Create a repository with " in description, like this one: https://github.com/ozpau/nbdev_new_cant_handle_this
2. Clone, cd into it, and call `nbdev_new`
3. This will generate invalid `index.ipynb`: https://github.com/ozpau/nbdev_new_cant_handle_this/blob/main/nbs/index.ipynb?short_path=7a1e639#L19

Notice that double quotation marks that breaks the json.

ozpau:
Fixes #1493

### DIFF
--- a/nbs/api/13_cli.ipynb
+++ b/nbs/api/13_cli.ipynb
@@ -52,7 +52,7 @@
     "\n",
     "from urllib.error import HTTPError\n",
     "from contextlib import redirect_stdout\n",
-    "import os, tarfile, sys"
+    "import os, tarfile, sys, json"
    ]
   },
   {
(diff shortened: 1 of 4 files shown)


nbs/api/13_cli.ipynb
```diff
@@ -372,7 +372,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "id": "b50d27fb-c343-47e2-a008-7ebdd02e1356",
```
ozpau, line 375: not sure why `nbdev_clean` didn't pick this one up

hamelsmu:
@jph00 I am not sure about this one, do we want to sanitize double quotes etc?

jph00:
> @jph00 I am not sure about this one, do we want to sanitize double quotes etc?

tbh I never use that CLI command, and I didn't write it, so I don't know! Sounds like maybe it is required for it to work correctly? Maybe you could give this PR a quick test to confirm it works OK for you?

PR #1497, opened 2025-02-21
feat: auto-generate contribute.md from nbs

bclavie:
This reproduces the same behaviour as for README.md, but for CONTRIBUTING.md, so every important root-level markdown file can be quarto-generated from notebooks.

Also puts our money where our mouth is, and replaces the manual `CONTRIBUTING.md` with the auto-generated one.


jph00:
Thanks @bclavie :) I'm happy to merge this PR, although I don't understand the point of it. Why use ipynb for this? If we don't have a good reason to suggest it, perhaps we should leave nbdev's as a .md?

bclavie:
Hey, one of my main motivation was to actually have a good reason to dive into `nbdev`, the second one (correct me if I'm wrong @audreyfeldroy) is in the name of uniformisation so things are easier to automate. I've got no strong opinion against reverting so that this repo (or any repo that doesn't really leverage quarto/doesn't need the automation, really) doesn't need to use it and just uses plain ol' `.md` though!

jph00:
Yeah let's keep the feature, but revert the use of it for the .md file here. AFAIK there isn't a benefit to auto-generating it, unless you need notebook features in it for some reason.

nbdev/quarto.py
```diff
@@ -259,14 +260,66 @@ def nbdev_readme(
...
+        if assets_folder.exists():
+            _copytree(assets_folder, cfg.config_path / assets_folder.name)
+
+# %% ../nbs/api/14_quarto.ipynb
+@call_parse
+def nbdev_contributing(
+    path:str=None,  # Path to notebooks (defaults to nbs_path)
+    chk_time:bool=False  # Only build if out-of-date
+):
+    """
```
jph00, line 291: docstrings should be on a single line.

nbdev/quarto.py
```diff
@@ -259,14 +260,66 @@ def nbdev_readme(
...
+    contrib_nb_name = cfg.get('contributing_nb', 'contributing.ipynb')
+    contrib_nb_path = path / contrib_nb_name
+    
+    contrib_md = cfg.config_path / 'CONTRIBUTING.md'
+    
+    # If out of date check is requested, skip if up-to-date or missing
+    if chk_time and _doc_mtime_not_older(contrib_md, contrib_nb_path):
+        return
+    
+    # If there's no contributing notebook, do nothing
```
jph00, line 308: comments should not just repeat what the code already does - only add comments where there's something surprising or hard to understand

nbdev/quarto.py
```diff
@@ -259,14 +260,66 @@ def nbdev_readme(
...
+    path = Path(path) if path else cfg.nbs_path
+    
+    # Decide which notebook is your "contributing" NB (you can hardcode or add to settings.ini)
+    contrib_nb_name = cfg.get('contributing_nb', 'contributing.ipynb')
+    contrib_nb_path = path / contrib_nb_name
+    
+    contrib_md = cfg.config_path / 'CONTRIBUTING.md'
+    
+    # If out of date check is requested, skip if up-to-date or missing
+    if chk_time and _doc_mtime_not_older(contrib_md, contrib_nb_path):
```
jph00, line 305: single ideas should be on a single line (e.g in this case the `if` and the body)

nbdev/quarto.py
```diff
@@ -259,14 +260,66 @@ def nbdev_readme(
...
+    Create CONTRIBUTING.md from contributing_nb (defaults to 'contributing.ipynb' if present).
+    Skips if the file doesn't exist.
+    """
+    cfg = get_config()
+    path = Path(path) if path else cfg.nbs_path
+    
+    # Decide which notebook is your "contributing" NB (you can hardcode or add to settings.ini)
+    contrib_nb_name = cfg.get('contributing_nb', 'contributing.ipynb')
+    contrib_nb_path = path / contrib_nb_name
+
```
jph00, line 301: empty lines should only be added where needed to separate substantive sections

jph00 (changes requested):
Thanks @bclavie :D

Minor style requests to look at before we merge. (I haven't mentioned all of them - just an example of each)

bclavie:
Thanks! Updated to better match styling conventions.

jph00:
Super. BTW for generating release notes, we use "enhancement" and "bug" labels, rather than adding a string prefix to the PR title. (I've added that now.)

PR #1499, opened 2025-02-22
feat: plaintext support

lukastk:
Closes https://github.com/AnswerDotAI/nbdev/issues/1498

Adds support for exporting the following plaintext formats as if they were regular `.ipynb` files:

- `percent`
- `light`
- `sphinx`
- `myst`
- `pandoc`

See [this](https://jupytext.readthedocs.io/en/latest/formats-scripts.html) for details on the above plain-text notebook formats.

Running `nbdev_export` now converts, by default, files of extension `*.{pct.py,lgt.py,spx.py,myst.md,pandoc.md,ipynb}`. See [this](https://github.com/lukastk/nbdev/blob/feature/plaintext-support/tests/plaintext.ipynb) for a demonstration (and some documentation) on the new exporter.

The feature requires `jupytext` and `nbformat` as (optional) dependencies. They will only be required if users attempt to export non-ipynb files.

### DIFF
--- a/nbdev/doclinks.py
+++ b/nbdev/doclinks.py
@@ -132,7 +132,7 @@ def nbglob(path=None, skip_folder_re = '^[_.]', file_glob='*.ipynb', skip_file_r
 def nbglob_cli(
     path:str=None, # Path to notebooks
     symlinks:bool=False, # Follow symlinks?
-    file_glob:str|List[str]=['*.ipynb'] + [f"*.{ext}" for ext in plaintext_file_formats], # Only include files matching glob
+    file_glob:Union[str, List[str]]=['*.ipynb'] + [f"*.{ext}" for ext in plaintext_file_formats], # Only include files matching glob
     file_re:str=None, # Only include files matching regex
     folder_re:str=None, # Only enter folders matching regex
     skip_file_glob:str=None, # Skip files matching glob
--- a/nbs/api/05_doclinks.ipynb
+++ b/nbs/api/05_doclinks.ipynb
@@ -344,7 +344,7 @@
     "def nbglob_cli(\n",
     "    path:str=None, # Path to notebooks\n",
     "    symlinks:bool=False, # Follow symlinks?\n",
-    "    file_glob:str|List[str]=['*.ipynb'] + [f\"*.{ext}\" for ext in plaintext_file_formats], # Only include files matching glob\n",
+    "    file_glob:Union[str, List[str]]=['*.ipynb'] + [f\"*.{ext}\" for ext in plaintext_file_formats], # Only include files matching glob\n",
     "    file_re:str=None, # Only include files matching regex\n",
     "    folder_re:str=None, # Only enter folders matching regex\n",
     "    skip_file_glob:str=None, # Skip files matching glob\n",


lukastk:
Note: In [this](https://github.com/AnswerDotAI/nbdev/pull/1499/commits/a235323667fee920b8d412002e4eccb9b394f378) commit I change the default behaviour of `nbdev_export` to also include the plaintext formats by default. This can be rolled back in case it's deemed too drastic a change. In which case, plaintext formats can be export by running, for example:

```bash
nbdev_export --file_glob '*.pct.py'
```

hamelsmu:
Thanks for providing the blog post with the motivation for this PR.  

Plain text formats don't capture outputs which are required to render nbdev docs so I'm afraid this is not a good fit.  Please let me know if I'm wrong or I am overlooking something!

lukastk:
Hi @hamelsmu. Thanks for taking the time to look through the PR and issue.

I agree that plaintext formats do not capture outputs, but I don't think these are *required* as such to render them via Quarto, which itself has support for rendering notebook-formatted `.py` files (see [this](https://quarto.org/docs/computations/render-scripts.html)).

I take your point that we often want to render the outputs in the docs as well, although for some use-cases (like mine) it's less necessary. One option would be to add a pre-processing step where the plaintext files are converted to `.ipynb` notebooks, and then executed. This way the outputs of the code cells would be rendered. I'd be happy to take a look into this if you think it's a good idea.

nbdev/doclinks.py
```diff
@@ -116,15 +117,18 @@ def nbglob(path=None, skip_folder_re = '^[_.]', file_glob='*.ipynb', skip_file_r
     "Find all files in a directory matching an extension given a config key."
     path = Path(path or get_config()[key])
     recursive=get_config().recursive
-    res = globtastic(path, file_glob=file_glob, skip_folder_re=skip_folder_re,
-                     skip_file_re=skip_file_re, recursive=recursive, **kwargs)
+    if type(file_glob) != list: file_glob = [file_glob]
+    res = []
```
jph00, line 125: This can be a list comprehension.

nbdev/doclinks.py
```diff
@@ -138,7 +142,7 @@ def nbglob_cli(
 @call_parse
 @delegates(nbglob_cli)
 def nbdev_export(
-    path:str=None, # Path or filename
+    path:str=None, # Path or filename,
```
jph00, line 149: Looks like a mistake?

nbdev/process.py
```diff
@@ -16,6 +16,14 @@
 
 from collections import defaultdict
 
+try:
+    import nbformat
```
jph00, line 20: imports can be on one line

nbdev/process.py
```diff
@@ -16,6 +16,14 @@
...
+    import jupytext
+    import tempfile
+    plaintext_supported = True
+except ImportError:
+    plaintext_supported = False
```
jph00, line 25:
```suggestion
except ImportError: plaintext_supported = False
```

nbdev/process.py
```diff
@@ -88,17 +96,43 @@ def _mk_procs(procs, nb): return L(procs).map(instantiate, nb=nb)
 # %% ../nbs/api/03_process.ipynb
 def _is_direc(f): return getattr(f, '__name__', '-')[-1]=='_'
 
+# %% ../nbs/api/03_process.ipynb
+plaintext_file_formats = {
```
jph00, line 100: This can be on one line.

nbdev/process.py
```diff
@@ -88,17 +96,43 @@ def _mk_procs(procs, nb): return L(procs).map(instantiate, nb=nb)
...
+            fmt = plaintext_file_formats[".".join(path.rsplit('.', 2)[-2:])] if fmt is None else fmt
+            if fmt in plaintext_file_formats.values():
+                if not plaintext_supported:
+                    raise ValueError(f"File {path} has a supported extension, but plaintext conversion is not supported. Please install jupytext and nbformat to use this feature.")
+                nb_converted = jupytext.read(path, fmt=fmt)
+                with tempfile.NamedTemporaryFile(delete=True, suffix=".ipynb") as temp_file:
+                    nbformat.write(nb_converted, temp_file.name)
+                    self.nb = read_nb(temp_file.name) if nb is None else nb
+                    return
+        if fmt is None or fmt == "ipynb":
```
jph00, line 131: Each part here can be on one line.

jph00 (changes requested):
Thank you for this PR. Let's make this opt in, disabled by default. You can add a setting for it. I've added some minor formatting requests.

It's unusual to have a separate "tests" notebook. Can this be integrated into the docs instead, creating tests that also illustrate the behavior?

lukastk:
Thanks @jph00 for your comments.

>  Let's make this opt in, disabled by default. You can add a setting for it. I've added some minor formatting requests.

Sounds good, will do!

> It's unusual to have a separate "tests" notebook. Can this be integrated into the docs instead, creating tests that also illustrate the behavior?

Sure, can integrate these into the docs.

jph00:
Looks like this is abandoned so closing for now.

PR #1503, opened 2025-03-03
Use plum for type dispatch

RensDimmendaal:
Part of moving dispatch and transform from fastcore to fasttransform (see [here](https://github.com/AnswerDotAI/fasttransform/issues/3).

This PR removes the use of `fastcore.dispatch.TypeDispatch` and replaces it with `plum.Function`.

This needs to be done before merging [the PR](https://github.com/AnswerDotAI/fastcore/pull/669) that removes dispatch and transform from fastcore. Because as long as nbdev depends on `fastcore.dispatch` removing the module will cause [fastcore's integration tests to break](https://github.com/AnswerDotAI/fastcore/actions/runs/13637679980/job/38120313432?pr=669).

Open questions:
- [ ] version bump to which version?
- [ ] changelog update?


RensDimmendaal:
will re-open w cleaned up git history.

PR #1513, opened 2025-03-25
Update settings.ini

benoit-cty:
Switch to `main` branch in `settings.ini` in the hope to solve https://github.com/AnswerDotAI/nbdev/issues/1509


jph00:
Looks like we need nbdev_export here. I'll make the change at my end. Thanks for the PR!

PR #1531, opened 2025-09-13
fix: duplication of directives with nbdev_update #1530

### LINKED ISSUE #1530: nbdev_update duplicates directives
When running `nbdev_update` some of the directives will be duplicated in the notebook (I haven't identified the pattern which will be duplicated and which not)
Therefore, a subsequent `nbdev_export` can result in a py file with the same directive appearing 3 times. 

I believe I have solved the issue and will open a  pull request.

daphn3k:
fixes #1530 

The actual fix is a couple of lines in `sync.ipynb` 's function `_update_nb()`. 
Below it, I have added two cells for comparing nbdev_update with and without the fix. 
To be able to compare (as well as reproducing the bug) I had to add a new notebook (tests/update_directives.ipynb) and its exported py file (nbdev/test_update_directives.py). 

This is my first pull request to this repo, so apologies if it doesn't fully conform. Feedback is welcome :)


jph00:
Many thanks for this PR! :)

tbh, it's way too much code. My basic premise is that I hate code. So every line of it must be justifiable by showing how much values it's adding. There are a lot of lines here not earning their keep. ;)

Instead, try to have just the absolute minimum about of new code to implement your feature, and to show it working. Don't add new files -- I add files of code even more than lines of code!

PR #1582, opened 2026-02-27
Fix issue #1581: _re_proj in update_proj replaces **all** name attributes in pyproject.toml

JonasJakubassa:
Fix issue https://github.com/AnswerDotAI/nbdev/issues/1581. See issue for description. 

Work done together with @janhesse53.

See also solveit dialog: https://share.solve.it.com/d/772cd0fefa8435ad2e3d26b4e43d5ce6

### DIFF
--- a/nbdev/config.py
+++ b/nbdev/config.py
@@ -267,11 +267,16 @@ dynamic = [ "keywords", "description", "version", "dependencies", "optional-depe
 
 [tool.uv]
 cache-keys = [{ file = "pyproject.toml" }, { file = "setup.py" }]
+
+[[tool.uv.index]]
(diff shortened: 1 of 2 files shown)


nbdev/config.py
```diff
@@ -267,11 +267,16 @@ def show_src(src, lang='python'): return Markdown(f'```{lang}\n{src}\n```')
 
 [tool.uv]
 cache-keys = [{ file = "pyproject.toml" }, { file = "setup.py" }]
+
```
jph00, line 270: I'll remove this - it's not something that should be default afaict.

jph00:
Thanks! :)

PR #1590, opened 2026-04-21
Improve error message raised when exporting relative import in notebook.

### LINKED ISSUE #1589: Better error message for relative imports when exporting notebooks
Currently when exporting a notebook with a relative import like:

```python
from .foo import bar
```

nbdev export fails with just a AssertionError that comes from `assert not level`

It would be better to raise a clear error message saying that relative imports are not supported during export. That would make the problem much easier to understand.

dienhoa:
Fixes https://github.com/AnswerDotAI/nbdev/issues/1589

Before this change, export failed with just `assert not level` when exporting with relative import, which did not explain the actual problem. This update tells the user that relative imports are not supported during export and includes the relevant context.

@jph00


dienhoa:
Hmm I'm not sure why the CI is failing. The notebook and the exported .py both seem clean

jph00:
Thanks! I'm not sure what's up with the CI either...

PR #1597, opened 2026-07-02
Set __file__ to the exported module path when testing notebooks

ncoop57:
## Summary

Sets `__file__` in the notebook execution namespace when `nbdev-test` runs an exported notebook, pointing at the module the notebook exports to. Path-dependent code like `Path(__file__).parent` now behaves the same under test as it does in the exported module.

Requires `fastcore>=1.14.4`, the release containing AnswerDotAI/fastcore#835 (now merged).

## Changes

- **Inject `__file__` before running cells**: if a notebook declares a `default_exp`, `test_nb` sets `__file__` in `CaptureShell.user_ns` to the exported module's path before `run_all`.
- **New fixture `tests/file_var.ipynb`**: asserts `__file__` resolves to `nbdev/file_var.py` and that an explicit assignment overrides the injected value; exercised from `nbs/api/12_test.ipynb`.

## Motivation

Exported modules commonly locate resources relative to themselves via `__file__`, which doesn't exist in the test namespace, so such code can't be exercised by notebook tests without workarounds. The full suite passing is meaningful coverage here: every notebook in `nbs/api` declares a `default_exp`, so they all now run with `__file__` set.


ncoop57:
~~Failures are due to an issue in linkify. PR fix: https://github.com/AnswerDotAI/nbdev/pull/1602~~ Already fixed in main

ncoop57:
@jph00 errors are due to the newest version of fastcore requiring python 3.11. Should we drop 3.10 from the tests or pin nbdev to a lower version of fastcore?

jph00:
Oh yeah good point - lemme look into this.

PR #1602, opened 2026-07-08
Fix linkify to handle list-form markdown outputs

ncoop57:
## Summary

Fixes the docs build (`nbdev-docs` / `nbdev-proc-nbs`), currently broken on main: `NbdevLookup.linkify` assumes markdown is a string, but notebooks store multiline mime data as lists of lines per the nbformat convention, so `add_links` crashes with `AttributeError: 'list' object has no attribute 'splitlines'` on any notebook with a saved markdown output.

## Changes

- **Join list input in `linkify`**: `if isinstance(md,list): md = ''.join(md)` before splitting into lines. The write side re-splits on save, so round-tripping is unchanged.
- **Test**: `linkify` on a list of lines returns the joined, linkified string.

## Motivation

Recent fastcore releases write notebooks with nbformat-standard split lines (AnswerDotAI/fastcore#836), so cleaned notebooks now store `text/markdown` outputs as lists, which `linkify` had never seen before. This surfaced in PR CI (the docs build check runs on PRs, e.g. #1597) but reproduces on main directly with `nbdev-proc-nbs`.

### DIFF
--- a/nbdev/doclinks.py
+++ b/nbdev/doclinks.py
@@ -296,6 +296,7 @@ class NbdevLookup:
 
     def linkify(self, md):
         if md:
+            if isinstance(md,list): md = ''.join(md)
             in_fence=False
             lines = md.splitlines()
             for i,l in enumerate(lines):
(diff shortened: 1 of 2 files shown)


ncoop57:
Superseded by f73d881 (fixes #1603)

PR #1620, opened 2026-07-29
Fix failing tests: add missing tempfile import in 04_export

RensDimmendaal:
The test cell 
in 04_export used `tempfile.TemporaryDirectory()` but never imported `tempfile`, causing failures. 
Adds the import alongside `warnings` at the top of the hide block and removes the now-redundant 
duplicate `import warnings`.


RensDimmendaal:
Closing, fixed directly on main.

